Issue #46, opened 2014-11-14
${PACKAGE_NAME}Config.cmake files generated in the build tree are broken


bartlettroscoe:
The files `${PACKAGE_NAME}Config.cmake` generated in the build tree:

```
${${PACKAGE_NAME}_BINARY_DIR}/${PACKAGE_NAME}Config.cmake
```

are broken.  That is, if a client CMake project tries to include these files, they are broken.

I will put in an automated test into TriBITS to demonstrate the problem.

bartlettroscoe:
This has been addressed by Brian's changes that were merged in pull request #54.

I am going to close this issue as fixed.

PR #54, opened 2015-01-28
Buildtree config cmake 46

briadam:
Proposed fix for TriBITS #46

This isn't perfect, but a big step up for build tree ProjectConfig.cmake and PackageConfig.cmake files.  Fixes issue with missing quote in the generated CMake code, properly generates includes from one PackageConfig.cmake to another, and puts all targets in a top-level build tree Targets.cmake.  Ideally these would be in separate PackageTargets.cmake in the build tree as is done for install, but I am not sufficiently TriBITS savvy to do this.

This should not affect install tree *Targets.cmake or *Config.cmake, but further testing is probably warranted.


briadam:
I tested this with Trilinos and VERa, but didn't verify the install *.cmake files.

bartlettroscoe:
Thanks Brian.  I looked over this and the changes were very minimal.  It passed my test for this (that I saw you merged in) and it also passed the test that ensures the installed <Package>Config.cmake files also work.

This has been merged to github/master as well as casl-dev/master.

PR #59, opened 2015-04-07
use FIND_PACKAGE(HDF5)

nschloe:


nschloe:
I just tested this with several versions of Debian and Ubuntu and this patch indeed makes finding the HDF5 libraries more robust. Specifically, TriBits doesn't work for Ubuntu Vivid and beyond since the libraries were slightly renamed here.

bartlettroscoe:
Nico,

As described in #67, we need to make some changes to the way that updated `FindTPL<TPLNAME>.cmake` modules behave.  In addition, to we will need to add arguments to `FIND_PACKAGE(HDF5)` to also find `hdf5_fortran` when Fortran is enabled. That will be needed to maintain backward compatibility, right?

bartlettroscoe:
Nico, look at the branch `findpackage-hdf5` in my git repo.  That should address all of the issues mentioned in #67.  I have tested this with Trilinos and CASL VERA and it works as it should.  That is, by default, if you don't specify anything, it will call `FIND_PACKAGE(HDF5 ...)` and it will look for `hdf5_fortran` if requested.

If I don't hear from you pretty quick, I will go ahead and push this.

bartlettroscoe:
I have pushed the updated `FindTPLHDF5.cmake` file to the master branch and snapshotted to Trilinos.  See #67 for details.

I am giving this to you to review @nschloe.

nschloe:
Closing this in favor of commit 2e44739d9ae8c073ab08d82cf14e4390888577de.

Issue #63, opened 2015-04-14
Merge TriBITS concepts and implementation of Packages and TPLs


bartlettroscoe:
The current design of TriBITS has TriBITS Packages and TriBITS TPLs and distinct separate entities.  TriBITS Packages are always built from source in the current TriBITS design while TriBITS TPLs are always pre-built and just pointed to at configure time.  This has been a successful model for Trilinos, CASL VERA, and other related projects for several years.  However, this model does not scale well to larger meta-projects and does not provide sufficient flexibility to handle more efficient alternative development workflows and other use cases.

The basic idea being proposed in this story is to extend TriBITS to allow some pre-built (and possibly installed) TriBITS packages to be treated as TPLs and then configure/build some downstream TriBITS packages from source.  This would be accomplished through the usage of `<Package>Config.cmake` files to get the needed information.  Likewise, this work would allow some TPLs could be built from inside of the TriBITS project like other TriBITS packages.

The details of the of the plan, the motivating use cases, and the design implications are given in [this Google Doc](https://docs.google.com/document/d/1T9KKrtgFm8-TSEuq119gdDR9tJMlBN3nAVjcN15Ge6k/edit?usp=sharing).

nschloe:
If I may comment on this from the perspective of a (distro) package maintainer:

Every piece of software has a bunch of dependencies, in former times only listed in a README, and more recently rigorously checked at configure time.
Some projects ship third-party software (for example [gmsh](https://github.com/live-clones/gmsh/tree/master/contrib) with the idea to make it easier for users to install the software. Unfortunately, in all cases I'm aware of, the bundled software is barely updated and doesn't encourage pushing changes upstream. This way, factual forks of the dependent libraries are created.

Constrasting this, PETSc does _not_ bundle its dependencies, but has its (custom) build system download them for you from the appropriate source. From a package maintainer's point of view, it's not quite clear why the fetch-&-build of TPL should be part of the build system for the actual code. Because, in fact, what you're doing is the job of a distribution (dependency managment, installation etc.).

A cleaner and equally easy way helping the user install TPLs is to provide a _separate_ script that installs all deps (e.g., `install_trilinos_dependencies.sh`). This way, the lazy user can use this convenience script, or install the dependencies via the distribution (e.g., `apt-get install libsuperlu-dev libnetcdf-dev [...]`). FEniCS goes that way.
The advantage that I see in this approach is a separation of concerns and consequently an easier-to-maintain codebase.

[...]

PR #75, opened 2015-06-29
export more system/compiler info

nschloe:
As requested by @gahansen.

### DIFF
--- a/tribits/core/installation/TribitsProjectConfigTemplate.cmake.in
+++ b/tribits/core/installation/TribitsProjectConfigTemplate.cmake.in
@@ -165,6 +165,15 @@ SET(${PROJECT_NAME}_MPI_EXEC_MAX_NUMPROCS "${MPI_EXEC_MAX_NUMPROCS}")
 SET(${PROJECT_NAME}_MPI_EXEC_NUMPROCS_FLAG "${MPI_EXEC_NUMPROCS_FLAG}")
 
 ## ---------------------------------------------------------------------------
+## Compiler vendor identifications
+## ---------------------------------------------------------------------------
+SET(${PROJECT_NAME}_SYSTEM_NAME "${CMAKE_SYSTEM_NAME}")
+SET(${PROJECT_NAME}_CXX_COMPILER_ID "${CMAKE_CXX_COMPILER_ID}")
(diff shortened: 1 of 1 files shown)


nschloe:
I believe this is ready for merging. :)

bartlettroscoe:
It would be good if we could add some checks to the automated tests to make sure this information is being set correctly.  It may be hard to write portable tests for the contents of these variables but we can check to make sure they are there.  We need to move TriBITS closer to [Self Sustaining Software](http://trac.trilinos.org/wiki/TribitsLifecycleModelOverview#self_sustaining_software) and that means we need to be adding automated tests as we add new features.  In can point out where to add checks for these in the test suite.

nschloe:
> I can point out where to add checks for these in the test suite.

Yes, please.

bartlettroscoe:
Put in `MESSAGE()` commands to print these new variables in:

   TriBITS/test/core/ExamplesUnitTests/DummyPackageClientCMakeLists.txt

and then put in grep statements for these variables in:

    TriBITS/test/core/ExamplesUnitTests/CMakeLists.txt

(just grep for `RunDummyPackageClientBulid.cmake` and you will see the other regex expressions).

For these tests to be portable, you will likely not be able to inspect the contents of these variables, just that they are present in the export file.  But at least that will stop someone from accidentally deleting these and the automated tests not noticing.  Add these tests as a new commit, then revert the commit that added the variables to the XXXConfig.cmake file and confirm the tests fail!

Writing stronger tests that look at the contents of these variables would require us too set up a mock env with mock compilers so they are the same on every platform.  There are other options too but the above "existence" test is the bare minimum test.  I am okay with an existence tests for now.

In general, if you are adding new features or extending the code and never add new tests or extend existing tests, that is a big red flag.   Don't you agree?

Make sense?

bartlettroscoe:
Wait, this is the project-level file, not the package-level file?  I don't think the project-level file is under any automated testing in TriBITS. Also, I am not sure that a project-level export file is going to be used in the refactoring in #63 at all.  So I guess since there is no testing for this file now and I their may not be a long-term future for this file (if #63 goes well), then I guess I can go ahead and merge into TriBITS, untested, since it is not going to break anything related to any of my projects.

nschloe:
> Wait, this is the project-level file, not the package-level file?

Indeed. @gahansen would certainly be glad about a merge.

Issue #77, opened 2015-07-15
Cannot find GIT in Windows


bkochuna:
In Windows TriBITS is unable to locate the git executable. This causes a configure failure when the generation of <PROJECT>RepoVersion.txt is enabled.

EDIT: The git format-patch commit is below.

bkochuna:
From fee7109f249ec897446377c307cbfa6bf91aa63c Mon Sep 17 00:00:00 2001
From: Brendan Kochunas <bkochuna@umich.edu>
Date: Wed, 15 Jul 2015 13:26:03 -0400
Subject: [PATCH] Fixing issue with finding git in Windows

---
 tribits/core/package_arch/TribitsConstants.cmake   |   11 -----------
 .../core/package_arch/TribitsGlobalMacros.cmake    |   15 +++++++++++++--
 2 files changed, 13 insertions(+), 13 deletions(-)

diff --git a/tribits/core/package_arch/TribitsConstants.cmake b/tribits/core/package_arch/TribitsConstants.cmake
index b628bf4..a4fc902 100644
--- a/tribits/core/package_arch/TribitsConstants.cmake
+++ b/tribits/core/package_arch/TribitsConstants.cmake
@@ -77,14 +77,3 @@ SET(${PROJECT_NAME}_CDASH_SUBPROJECT_DEPS_XML_FILE_NAME CDashSubprojectDependenc
 SET(${PROJECT_NAME}_PACKAGE_DEPS_TABLE_HTML_FILE_NAME ${PROJECT_NAME}PackageDependenciesTable.html)

 SET(${PROJECT_NAME}_PACKAGE_DEPS_FILES_DIR "cmake/dependencies")
-
-# Other stuff
-
-IF(WIN32)
-  #Apparently FIND_PROGRAM looks for an exact match of the file name.
-  #So even though "git clone ..." is valid to use on windows we need to give the
-  #full name of the command we want to run.
-  SET(GIT_NAME git.cmd)
-ELSE(WIN32)
-  SET(GIT_NAME git)
-ENDIF(WIN32)
diff --git a/tribits/core/package_arch/TribitsGlobalMacros.cmake b/tribits/core/package_arch/TribitsGlobalMacros.cmake
index 09afbc5..1f9b0c5 100644
--- a/tribits/core/package_arch/TribitsGlobalMacros.cmake
+++ b/tribits/core/package_arch/TribitsGlobalMacros.cmake
[...]

PR #78, opened 2015-07-15
Fixing issue in finding git on windows (TriBITS #77)

bkochuna:
This patch fixes the issue described in #77


bartlettroscoe:
Brendan,

I think there must be Windows systems where git is called 'git.cmd'.  I bet this is the machine where Trilinos Windows testing runs.  Therefore, I suspect your change will break those systems.  Also, the variable `GIT_NAME` is not only used in the file `TribitsGlobalMacros.cmake`.  It is also used in `TribitsCTestDriverCore.cmake`.  I am sure this is what broke the TriBITS test suite shown [here](https://travis-ci.org/TriBITSPub/TriBITS/builds/71120243).

I think what is needed is to instead define a new function called `TRIBITS_FIND_GIT()` and have it find git and return a variable `GIT_PATH` with the absolute path to the found git (or return `NOTFOUND`).  Then this function can be used in `TribitsGlobalMacros.cmake` and `TribitsCTestDriverCore.cmake` (or any CMake code that wants to find git).  Does that make sense?

bkochuna:
Ross,

Ok, thanks for the feedback. I'll tweek the modification as you suggest and issue a new pull request. I don't know when I'll have time to get to that as the change is getting a little more involved. So, sit tight till you see a new pull request.

Thanks!

PR #82, opened 2015-08-27
Fixed typos only.

jwillenbring:
Victoria Smith pointed out some typos in the TrilinosBuildReference that originate in TriBITS source code.  I also corrected other identical mis-spellings.


bartlettroscoe:
Thanks.  I will have a look and integrate and snapshot back to Trilinos.

PR #88, opened 2015-10-17
travis: switch to trusty, remove custom cmake ppa

nschloe:


bartlettroscoe:
I can only test this by actually pushing the commits so I did.  Let's see what Travis does with this.

nschloe:
PRs are tested automatically, aren't they?

This can be adjusted in travis, see the settings page for tribits over there.

Also, in GitHub, you will probably protect master such that nobody can push to it and PRs can only be merged if the tests pass.

bartlettroscoe:
That is interesting.  Will .travis on a branch be tested with Travis CI?  That is changing the definition of testing really in an acceptable way.  Therefore, someone could gut .travis in their PR so that it always passes, and then allow the PR to be accepted, essentially with no testing.  That seems wrong.

nschloe:
I'm not sure I understand. Only branches in the same repo and PRs are tested. See the x and checkmark at

h

above. I could of course have edited `.travis.yml` to remove all tests, but hopefully you would have caught that.

bartlettroscoe:
Looks like it ran the tests:

   https://travis-ci.org/TriBITSPub/TriBITS/builds/86221209

I am going to get this set up to post to the CDash site:

    http://my.cdash.org/index.php?project=TriBITS

Hopefully that will not be too hard.

Issue #100, opened 2015-12-04
gitdist: Fails for case of ambiguous 'HEAD'


cgbaker:
Warnings from git are not well handled by `gitdist dist-repo-status`. 
For example, 
```
$ git tag HEAD
$ git status
warning: refname 'HEAD' is ambiguous.
warning: refname 'HEAD' is ambiguous.
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean
```
will put my repo in a state where git commands issue warnings. 

Some commands, such as the one used to determine the local branch, also result in errors:
```
git  rev-parse --abbrev-ref HEAD
warning: refname 'HEAD' is ambiguous.
error: refname 'HEAD' is ambiguous
```

Then, if I attempt a repo status, the parsing crashes: 
```
$ gitdist dist-repo-status
Traceback (most recent call last):
  File "/usr/local/bin/gitdist", line 994, in <module>
    extraRepoStats = getRepoStats(options)
  File "/usr/local/bin/gitdist", line 873, in getRepoStats
    numCommits = getNumCommitsWrtTrackingBranch(options, trackingBranch, getCmndOutputFunc)
  File "/usr/local/bin/gitdist", line 785, in getNumCommitsWrtTrackingBranch
    numAuthorCommits = int(summaryLine.strip().split()[0].strip())
ValueError: invalid literal for int() with base 10: 'warning:'
```

The workaround, if desirable, is to filter out these lines where possible. I'm preparing a pull request with some changes illustrating some workarounds.

bartlettroscoe:
@cgbaker, what version of Python and git are you using?  I run gitdist on repos all the time that are in a deteched head state and not on tracking branches and it works just fine.

bartlettroscoe:
Okay, I have reproduced this with git 2.1.0:

```
[8vt@u235 VERA (master)]$ cp -r TeuchosWrappersExt TeuchosWrappersExt.copy
[8vt@u235 VERA (master)]$ cd TeuchosWrappersExt.copy/
[8vt@u235 TeuchosWrappersExt.copy (master)]$ git tag HEAD
[8vt@u235 TeuchosWrappersExt.copy (master)]$ git  rev-parse --abbrev-ref HEAD
warning: refname 'HEAD' is ambiguous.
error: refname 'HEAD' is ambiguous
```
I should be able to add a unit test for this strange git repo setup and then apply your fix.

cgbaker:
[...]

PR #101, opened 2015-12-04
filters warnings and errors in some git responses per #100

cgbaker:
see #100


bartlettroscoe:
@cgbaker, I rebased your commit on top of a new failing unit test commit and then made a small change as described in commit 3102a4f.  Therefore, I applied your change but since I did not directly merge your branch, it did not close this PR.  Therefore, I am closing it.

PR #123, opened 2016-05-26
If NetCDF is also enabled, may need hdf5_hl library

gsjaardema:
If using the netcdf4-enabled version of NetCDF library, it requires both libhdf5_hl and libhdf5.  

Check whether NetCDF is enabled and if so, add libhdf5_hl to the list of required libs.  This check should possibly also check if netcdf4 enabled in netcdf, but currently assumes that is the case.

### DIFF
--- a/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/tribits/common_tpls/FindTPLHDF5.cmake
@@ -21,6 +21,10 @@ IF (TPL_ENABLE_MPI)
   SET(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
 ENDIF()
 
+IF (TPL_ENABLE_Netcdf)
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
Looks reasonable to me.  Note that CASL VERA, ATTBDevEnv.cmake (trilinos/Trilinos#172) and SEMSDevEnv.cmake (trilinos/Trilinos#158) all explicitly list out the HDF5 libs for these builds.  Therefore, this change only impacts the default search for HDF5 libs.

I will merge into TriBITS 'master'.  However, this will need to be [snapshotted into Trilinos](http://trac.trilinos.org/wiki/TriBITSTrilinosDev#update_snapshot) and the SEACAS repo for this to impact those projects.

bartlettroscoe:
Note that there are no automated tests in TriBITS for FindTPLHDF5.cmake so I will take your word for it that this works.  But it looks perfectly safe and again will not break behavior for configures the specify `HDF5_LIBRARY_NAMES`.

PR #124, opened 2016-06-02
Update configure.py

nschloe:
Discovered by Debian's linitian.

### DIFF
--- a/tribits/examples/TribitsExampleProject/packages/wrap_external/external_func/configure.py
+++ b/tribits/examples/TribitsExampleProject/packages/wrap_external/external_func/configure.py
@@ -1,4 +1,4 @@
-#!/bin/python
+#!/usr/bin/env python
 
 import sys
 import os


bartlettroscoe:
Yes, use `/usr/bin/env python` instead of `/bin/python`.  That tool [Lintian](https://en.wikipedia.org/wiki/Lintian) seems pretty useful.

I rebased this commit and pushed it as 5a641e282a48d9c6675ef3375ea273c6e050cd05.

So this script is used as part of the TribitsExampleProject and it used in automated testing of TriBITS itself (but that testing is never run as part of Trilinos anymore).

Is it okay if we don't apply this back to the Trilinos 12.6 branch?  Given that no one will be running this script from Trilinos, it can't cause any defects.   The next time TriBITS gets snapshotted into Triinos, this will be part of the next Trilinos release automatically.

bartlettroscoe:
So Travis CI says that my pushed commit 5a641e2 [failed](https://travis-ci.org/TriBITSPub/TriBITS/builds/134741010).   It showed the failure:

```
$ sudo apt-get update -qq

W: There is no public key available for the following key IDs:

1397BC53640DB551

W: Failed to fetch https://apt.dockerproject.org/repo/dists/ubuntu-trusty/main/binary-amd64/Packages  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.

The command "sudo apt-get update -qq" failed and exited with 100 during .

Your build has been stopped.
```

I hope this was just a fluke and will clear up on the next CI iteration.  But this fetch failure really seems to have nothing to do with this commit.

nschloe:
It does look like a fluke.

In any case, I guess one doesn't really _need_ to `apt update`, so perhaps scratching that will already help.

bartlettroscoe:
Since this was merged, am closing.

PR #128, opened 2016-06-08
prefer parallel hdf5 when using MPI

nschloe:
Cf. https://cmake.org/cmake/help/v3.4/module/FindHDF5.html.


bartlettroscoe:
Is this still not resolved?  Is this safe to apply?

@gsjaardema, is this the right thing to do for HDF5?  I don't have a horse in this race because my configure scripts always just manually set the libraries so I don't rely on the automatic default finds.

gsjaardema:
This should be ok, but I don't have any experience with it.  

I usually build my own HDF5 libraries and explicitly set the HDF5_ROOT to point to the serial or parallel version that I want.

nschloe:
Using `HDF5_ROOT` (or `HDF5_INCLUDE_DIRS`/`HDF5_LIBRARY_DIRS`) overrides the preference. This PR is just meant to provide a sane default.

PR #145, opened 2016-09-13
Update openmp flag logic for newer CMake versions

gsjaardema:
CMake as of at least 3.2.3 does set `OpenMP_Fortran_FLAGS`, so check whether that symbol is set and if it is, use it.

This is at least needed on some Mac systems since the `OpenMP_Fortran_FLAGS` are different than the `OpenMP_C_FLAGS` (`-fopenmp` vs `-fopenmp=libomp`)


bartlettroscoe:
@gsjaardema,

Note that the minimum CMake version for TriBITS and Trilinos is still CMake 2.8.11.  Therefore, we will need to put in a conditional if-statement before we can push a change like this.

What is the best way to test this for Trilinos to know that it is working correctly?

-Ross

gsjaardema:
The patch does have an if-statement

bartlettroscoe:
Got you.  I will enable OpenMP and test this out.  If it looks good with Trilinos, I will push.

bartlettroscoe:
Before pushing this commit to the main TriBITS repo 'master' branch, I tested this against TribitsExampleProject using CMake 3.5.1 and GCC 4.8.3.  I used the configure script:

```
#!/bin/bash -e

TRIBITS_DIR=$HOME/localhome/PROJECTS/Trilinos.base/Trilinos/TriBITS/tribits

cmake \
-DTPL_ENABLE_MPI=ON \
-DCMAKE_BUILD_TYPE=DEBUG \
-DTribitsExProj_TRIBITS_DIR=$TRIBITS_DIR \
-DTribitsExProj_ENABLE_OpenMP=ON \
-DTribitsExProj_ENABLE_ALL_PACKAGES=ON \
-DTribitsExProj_ENABLE_TESTS=ON \
$TRIBITS_DIR/examples/TribitsExampleProject
```

The build then showed the `-fopenmp` flag in the compilation of a Fortran file:

```
$ make VERBOSE=1
...
[ 28%] Building Fortran object packages/mixed_lang/src/CMakeFiles/mixedlang.dir/Ray_Tracer_Kernel.f90.o
...  -fopenmp -g -O0   ...
...
```

I commented out [the line[(https://github.com/TriBITSPub/TriBITS/pull/145/files#diff-a48586a82325c07d963592125f584749R1805):

```
SET(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${OpenMP_Fortran_FLAGS}")
```

and verified that `OpenMP_Fortran_FLAGS` is indeed set.

I will not push.

bartlettroscoe:
@gsjaardema,

Does this change in TriBITS need to be synced to Trilinos urgently or do you just need this for SEACAS for now?

gsjaardema:
I just need it in SEACAS for now.

Thanks, (I assume the "not push" was supposed to say  "now push" ?)
..Greg

bartlettroscoe:
> I assume the "not push" was supposed to say "now push"

Yes "now push".  I seem to do that a lot for some reason.

This change will get synced back to Trilinos at some point.

Thanks for taking care of this!

Issue #174, opened 2017-02-09
Add install of Anaconda2 to devtools install


bartlettroscoe:
This story is to consider the addition of the Anaconda2 library:

* https://www.continuum.io/downloads

to the devtools_install collection and to the install_devtools.py script.

This is needed to install updated version of Python on various systems.

bartlettroscoe:
@leerw-ornl-gov, I pushed your branch for adding the install of anaconda2 to the the TriBITS devtools_install to the branch:

* https://github.com/TriBITSPub/TriBITS/tree/add-anaconda-install-174

and I created the PR #175 to facilitate a code review.

bartlettroscoe:
@MisterT1958, note PR #175.  We will have discussion about this change there.

bartlettroscoe:
As discussed in #175, we will pursue a VERA-specific solution for the installation of Python 2.7 and Anacanda2 in:

* https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/4558

PR #175, opened 2017-02-10
Add anaconda2 to install to devtools_install

bartlettroscoe:
This PR is to allow the code review and manage the shared development of #174.

CC: @leerw-ornl-gov

### DIFF
--- a/tribits/devtools_install/install-anaconda2.py
+++ b/tribits/devtools_install/install-anaconda2.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python
+import os, sys, urllib2
+
+# @HEADER
+# ************************************************************************
+#
+#            TriBITS: Tribal Build, Integrate, and Test System
+#                    Copyright 2013 Sandia Corporation
+#
+# Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+# the U.S. Government retains certain rights in this software.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+# 1. Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the Corporation nor the names of the
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY SANDIA CORPORATION "AS IS" AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL SANDIA CORPORATION OR THE
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+# ************************************************************************
+# @HEADER
+
+
+#
+# Defaults
+#
+
+CONDA_BASE_NAME = 'anaconda2'
+CONDA_DEFAULT_VERSION = '4.3.0'
+CONDA_SUPPORTED_VERSIONS = [ '4.2.0', '4.3.0' ]
+CONDA_INSTALLER_VERSIONS = { '4.2.0': '4.2.0', '4.3.0' : '4.3.0' }
+
+
+#
+# Script code
+#
+
+
+from InstallProgramDriver import *
+from GeneralScriptSupport import *
+
+
+class Anaconda2Install:
+
+  def __init__( self ):
+    pass
+
+  #
+  # Called before even knowing the product version
+  #
+
+  def getScriptName( self ):
+    return "install-anaconda2.py"
+
+  def getProductBaseName( self ):
+    return CONDA_BASE_NAME
+
+  def getProductDefaultVersion( self ):
+    return CONDA_DEFAULT_VERSION
+
+  def getProductSupportedVersions(self):
+    return CONDA_SUPPORTED_VERSIONS
+
+  #
+  # Called after knowing the product version but before parsing the
+  # command-line.
+  #
+
+  def getProductName(self, version):
+    return CONDA_BASE_NAME + "-" + version
+
+  def getBaseDirName(self, version):
+    return CONDA_BASE_NAME + "-" + version + "-base"
+
+  def getExtraHelpStr(self, version):
+    return """
+This script installs """ + self.getProductName(version) + """.
+"""
+
+
+  def injectExtraCmndLineOptions(self, clp, version):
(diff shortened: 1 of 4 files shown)


bartlettroscoe:
@leerw-ornl-gov and @MisterT1958,

The new code in this PR code seems to be working to download scripts with the name `Anaconda2-<full_version>-Linux-x86_64.sh` and then just running it.  Could we not just manually download the script for the desired version and them mirror it along with the other sources in a repo under github.com/tribitsdevtools/?  That approach (the same approach used by all of the other tools that are installed) has some advantages:

1. We control when these scripts get updated and therefore we better ensure reproducability of old build envs.  (Also, if that original site goes away, we can still longer install older versions.)

2. We remove a lot of the more complex logic in the scripts install-anaconda2.py and install_devtools.py that works to do the download of `Anaconda2-<full_version>-Linux-x86_64.sh` .  (Cloning a git repo is simple.)

3. It maintains the instructions to obtain the access to the git repos as either a tarball of the git repos on DVD (requiring no internet access at all) or directly cloning from the casl-dev repos or github mirrors as described in the [VERA Installation Guide](https://vminfo.casl.gov/public/VERA/doc/installation_guide/VERAInstallationGuide.html#install-the-base-development-environment).  (What is currently implemented in this PR breaks that process by requiring general internet access.  We did a lot of work to provide those three sources for the dev env data to do the install.  There are customers that need each of these modes.)  

Also, what is the script `Anaconda2-<full_version>-Linux-x86_64.sh` actually doing?  I just downloaded one of these and it is huge:

```
$ du -sh Downloads/Anaconda2-4.3.0-Linux-x86_64.sh 
462M    Downloads/Anaconda2-4.3.0-Linux-x86_64.sh
```

Did they actually put all of the Python source code and/or binary into a single *.sh script and installing requires no data outside of this script at all?  That seems pretty crazy but be good because it would not require an extra internet access to do the install.

The license as described in the script is:

```
================
Anaconda License
================

Copyright 2016, Continuum Analytics, Inc.

All rights reserved under the 3-clause BSD License:

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:

* Redistributions of source code must retain the above copyright notice,
this list of conditions and the following disclaimer.

* Redistributions in binary form must reproduce the above copyright notice,
this list of conditions and the following disclaimer in the documentation
and/or other materials provided with the distribution.

* Neither the name of Continuum Analytics, Inc. nor the names of its
contributors may be used to endorse or promote products derived from this
software without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
ARE DISCLAIMED. IN NO EVENT SHALL CONTINUUM ANALYTICS, INC. BE LIABLE FOR
ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
DAMAGE.
```

That license would seem to suggest that we can copy these *.sh scripts and put them in our own git repos.

After you have thought about this some, can we have a conversation to discuss this?

Also, in the future, if changes are going to be made to TriBITS proper, then it is best to follow the process documented here:

* https://github.com/TriBITSPub/TriBITS/wiki/Contributing-to-TriBITS

That way we can have a high-level discussion like this before a lot of work is done implementing code.  (That just saves everyone a lot of time.)

leerw-ornl-gov:
First, the Anaconda install differs from the one pattern (download source, configure, build, install) that devtools_install inherently supports.  The Anaconda pattern is a very simple binary install.

> The new code in this PR code seems to be working to download scripts with the name Anaconda2-<full_version>-Linux-x86_64.sh and then just running it. Could we not just manually download the script for the desired version and them mirror it ...
>
>    We control when these scripts get updated and therefore we better ensure reproducability of old build envs. (Also, if that original site goes away, we can still longer install older versions.)

True.

>    We remove a lot of the more complex logic in the scripts install-anaconda2.py and install_devtools.py that works to do the download of Anaconda2-<full_version>-Linux-x86_64.sh . (Cloning a git repo is simple.)

The main reason is the file to download is too big.  Github rejects files larger than 100 MB.

Although somewhat irrelevant, fetching a URL is the easiest, most ReSTful network operation available-- just saying.

>    It maintains the instructions to obtain the access to the git repos as either a tarball of the git repos on DVD (requiring no internet access at all) or directly cloning from the casl-dev repos or github mirrors as described in the VERA Installation Guide. (What is currently implemented in this PR breaks that process by requiring general internet access. We did a lot of work to provide those three sources for the dev env data to do the install. There are customers that need each of these modes.)

I understand the need to have local repositories (in the general sense not specifically a Git repo), and binary install files can be written to a DVD.

Philosophizing for a moment...

If all that is needed is a network-accessible URL from which to download install files, especially large ones, Git is at best overkill and in practice barfs on large files.  Access via an FTP or HTTP(S) URL makes most sense, and all sorts of mechanisms exist to support it, such as running Apache or a local DropBox service.

  "If you only have a hammer, you tend to see every problem as a nail."

> Also, what is the script Anaconda2-<full_version>-Linux-x86_64.sh actually doing? I just downloaded one of these and it is huge:
> ...
> Did they actually put all of the Python source code and/or binary into a single *.sh script and installing requires no data outside of this script at all? That seems pretty crazy but be good because it would not require an extra internet access to do the install.

Just binary.  Many (if not most) tools offer binary installs.  Linux distro package managers  provide binary installs.

> The license as described in the script is:

Yes.  I checked the license to ensure it allows us to serve the installer and/or write it to a DVD.

> After you have thought about this some, can we have a conversation to discuss this?
>
> Also, in the future, if changes are going to be made to TriBITS proper, then it is best to follow the process documented here:
> 
>    https://github.com/TriBITSPub/TriBITS/wiki/Contributing-to-TriBITS
>
> That way we can have a high-level discussion like this before a lot of work is done implementing code. (That just saves everyone a lot of time.)

Understood.  This started as a VERA issue, which brings up another discussion outside of the domain of TriBITS.

The VERA install process offers only two mechanisms for installing tools into the environment: TriBITS devtools_install and TPLs.  We looked at installing Anaconda Python as a TPL, but this doesn't fit for things like updating the environment scripts.  Semantically, Python fits better as a "common tool", but devtools_install is inherently well suited for binary installs.

Perhaps the better approach is to augment the VERA install process and not mess with TriBITS.

bartlettroscoe:
@leerw-ornl-gov,

You points are well taken about the size of the Anaconda2 install script not fitting in GitHub.

Thinking about this more, a binary install that is pulled down directly from the web is pretty radically different from the other tools that are installed with the TriBITS devtools_install scripts.  The devtools_install scripts are designed to provide from-source builds on a number of machines and platforms and therefore be pretty portable (or as portable as the native build systems for those build tools will allow, GCC being the most difficult).  The process for the Anaconda2 install is radically different and does not really fit into the workflow for the existing devtool_install/ scripts.

I am thinking it might be easier and more general just to treat the install of Ananoda2 as its own tool in the VERA Installation Guide.  I think we could just update that document to provide instructions for downloading the binary install script and the command to run it (using the env vars defined earlier in that document).  Then we might just add an argument to install_devtools.py to prepend and arbitrary path that is loaded by the installed load_dev_env.sh script.   That way, we could manage all of the use cases for installing the VERA Dev Env in a more straightforward way.

I can outline this more in:

* https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/4558

Does that sound reasonable?  Let's discuss the details in that Trac ticket.

leerw-ornl-gov:
@bartlettroscoe

> Does that sound reasonable?

Yes.  Forcing this into devtools_install doesn't make a lot of sense-- square peg round hole.

> Let's discuss the details in that Trac ticket.

That's what I originally intended for https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/4558, before 4556 kind of subsumed it.  If you think we should create a new one I'll do so.

bartlettroscoe:
Okay, then we will close this PR and #174 and continue the VERA-specific discussion:

* https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/4558

Sorry about extra work you had to do to put together this Python install code.  However, I think you might still be able to to use most of it if you want a more automated process.  But a single download and a single script invocation may not be worth scripting up, just list it out in the install guide.

PR #197, opened 2017-06-14
Dev guide typos

fryeguy52:


bartlettroscoe:
@fryeguy52, I merged this PR branch and and pushed to the 'master' branch after making a few small changes (see the updated commits).

We should briefly talk through some of your questions in the document in the form "(*This is a question*)".  I moved those to RST comments so we can find them.

PR #202, opened 2017-07-06
Gitdist with Python3

jmgate:
Outputs from commands run via the shell needed to be wrapped in s() to
decode the byte strings into Unicode strings.

Previously, using gitdist with Python 3.5.2 yielded:
```
Traceback (most recent call last):
  File “./Trilinos/cmake/tribits/python_utils/gitdist”, line 1453, in <module>
    + addColorToRepoDir(options.useColor, repoName) )
  File “./Trilinos/cmake/tribits/python_utils/gitdist”, line 814, in addColorToRepoDir
    return txtbld+txtblu+strIn+txtrst
TypeError: can’t concat bytes to str
```
This fixes the problem.

### DIFF
--- a/tribits/python_utils/gitdist.py
+++ b/tribits/python_utils/gitdist.py
@@ -771,7 +771,7 @@ def filterWarnings(lines):
 def getCmndOutput(cmnd, rtnCode=False):
   child = subprocess.Popen(cmnd, shell=True, stdout=subprocess.PIPE,
     stderr = subprocess.STDOUT)
-  output = child.stdout.read()
+  output = s(child.stdout.read())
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
In parallel, I made exactly the same fix and already pushed in commit 681b0f386ca57bc2b35a31111d1900836ab544f0 before I saw this.  Also, I updated the unit tests for gitdist and they all now pass with Python 3.5.2.

Therefore, I will close this PR.  Sorry for the trouble.

jmgate:
Sounds good.  Good to know we wound up at the same solution.

PR #203, opened 2017-07-06
Run gitdist from anywhere in repo

jmgate:
Look at the current working directory when `gitdist` is executed, determine the top-level `git` repo in that path, and change to it before continuing.  This allows `gitdist` to be run from anywhere in a `git` repository, including within a nesting of repositories.  For instance, if you have a directory structure like the following
```
/some/path/to/wherever/
  metaRepo/
    .git/
    .gitdist
    repo1/
      .git
    repo2/
      .git
    etc.
```
executing `gitdist` anywhere down in the `metaRepo` directory tree will change you to the `metaRepo` directory, execute `gitdist`, and then leave you back where you started when `gitdist` finishes.

If, for whatever reason, this change is not a good idea, just let me know and I can simply write a shell script to wrap `gitdist` for my own personal use.

### DIFF
--- a/tribits/python_utils/gitdist.py
+++ b/tribits/python_utils/gitdist.py
@@ -1035,25 +1035,41 @@ def getCommandlineOps():
 
   # Change to top-level git directory (in case of nested git repos)
 
-  drive, currentPath = os.path.splitdrive(os.getcwd())
-  pathList = []
-  while 1:
-    currentPath, currentDir = os.path.split(currentPath)
-    if currentDir != "":
-      pathList.append(currentDir)
-    else:
-      if currentPath != "":
-        pathList.append(currentPath)
-      break
-  pathList.reverse()
-  newDir = drive+pathList[0]
-  for currentDir in pathList:
-    newDir = os.path.join(newDir, currentDir)
-    if os.path.isfile(os.path.join(newDir, ".gitdist")):
-      break
-    if os.path.isfile(os.path.join(newDir, ".gitdist.default")):
-      break
-  os.chdir(newDir)
+  moveToBaseDir = os.environ.get("GITDIST_MOVE_TO_BASE_DIR")
+
+  if moveToBaseDir == None or moveToBaseDir == "":
+    # Run gitdist in the current directory
+    None
+  elif moveToBaseDir == "EXTREME_BASE":
+    # Run gitdist in the most base dir where .gitdist[.default] exists
+    drive, currentPath = os.path.splitdrive(os.getcwd())
+    pathList = []
+    while 1:
+      currentPath, currentDir = os.path.split(currentPath)
+      if currentDir != "":
+        pathList.append(currentDir)
+      else:
+        if currentPath != "":
+          pathList.append(currentPath)
+        break
+    #print("pathList="+str(pathList))
+    pathList.reverse()
+    #print("pathList="+str(pathList))
+    newDir = drive+pathList[0]
+    for currentDir in pathList:
+      #print("currentDir="+str(currentDir))
+      newDir = os.path.join(newDir, currentDir)
+      #print("newDir="+str(newDir))
+      if os.path.isfile(os.path.join(newDir, ".gitdist")):
+        break
+      if os.path.isfile(os.path.join(newDir, ".gitdist.default")):
+        break
+    os.chdir(newDir)
+  else:
+    print(
+      "Error, env var GITDIST_MOVE_TO_BASE_DIR='"+moveToBaseDir+"' is invalid!"+
+      "  Valid choices include empty '', and EXTREME_BASE")
+    sys.exit(1)
 
   # Get list of extra repos
 
--- a/tribits/python_utils/gitdist.py
+++ b/tribits/python_utils/gitdist.py
@@ -1033,6 +1033,28 @@ def getCommandlineOps():
   # E) Get the list of extra repos
   #
 
+  # Change to top-level git directory (in case of nested git repos)
+
+  drive, currentPath = os.path.splitdrive(os.getcwd())
+  pathList = []
+  while 1:
+    currentPath, currentDir = os.path.split(currentPath)
+    if currentDir != "":
+      pathList.append(currentDir)
+    else:
+      if currentPath != "":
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
I checked out the branch and it seems to have broken the usage for me.  For example, the current version of gitdist shows:

```
[rabartl@crf450 Trilinos (develop)]$ gitdist dist-repo-status
---------------------------------------------------------------------
| ID | Repo Dir           | Branch  | Tracking Branch   | C | M | ? |
|----|--------------------|---------|-------------------|---|---|---|
|  0 | Trilinos (Base)    | develop | github/develop    |   |   |   |
|  1 | TriBITS            | master  | github/master     |   |   |   |
|  2 | TriBITS/TriBITSDoc | master  | bb/master         |   |   |   |
|  3 | kokkos             | master  | github/master     |   |   |   |
|  4 | seacas             | master  | gsjaardema/master |   |   |   |
|  5 | DrekarBase         | master  | ssg/master        |   |   |   |
|  6 | DrekarResearch     | master  | ssg/master        |   |   |   |
---------------------------------------------------------------------
```

But with the version on this branch, it shows:

```
[rabartl@crf450 Trilinos (develop)]$ gitdist dist-repo-status
---------------------------------------------------------------
| ID | Repo Dir       | Branch | Tracking Branch | C | M | ?  |
|----|----------------|--------|-----------------|---|---|----|
|  0 | rabartl (Base) | master | origin/master   |   | 1 | 13 |
---------------------------------------------------------------
```

The issue is that I use git to manage my home directory so when I run gitdist it will always just run in my home dir.

I think I know how to fix this so I will give it a shot based on your branch real quick.  But before we can push to TirBITS, we will need some automated tests to demonstrate and protect correct behavior.

jmgate:
It may be better for me to just write a little wrapper script for my own personal use case, which is totally fine.

bartlettroscoe:
> If I was somewhere in Trilinos, I'd want to run git commands (checkout, etc.) on Trilinos, tcad-charon, charon-data, etc., which would require going up to charonMetaRepo. If it stopped after finding the .gitdist.default in Trilinos, it would only execute them in the Trilinos repo, right? Then I would potentially be on a different branch in Trilinos compared to the other repos.

But that is the problem with changing this behavior.  Different people will want different behavior for different use cases.

We might be able to resolve this with some system-level preferences.  We might, for example, allow the definition of a `~/.gitdistconfig` file that would be read by gitdist that would allow an individual user to select how they want gitdist to behave.  For example, there might be an option in `~/.gitdistconfig` called `moveToBaseDirMode` and it could have options like `noMove` (the current behavior), `moveToImmediateBase` (the behavior I want) and `moveToExtremeBase` (the behavior that you want).

If this sounds like a good idea to you, we can create a new GitHub issue for gitdist to support a `.gitdistconfig` file and then a new GitHub issue for the allowing customization of the behavior of moving to subdirs as described above.

P.S. the updated commit I just pushed implements the mode `moveToBaseDirMode = moveToExtremeBase`.

bartlettroscoe:
@jmgate,

I just pushed the commit 82822cb to your branch for this PR which causes gitdist to read an env var `GITDIST_MOVE_TO_BASE_DIR`.  If that env var is set to `EXTREME_BASE`, then I think the script has the behavior your want.  Please pull this branch and give it a try with:

```
$ export GITDIST_MOVE_TO_BASE_DIR=EXTREME_BASE
$ cd charonMetaRepo/Trilinos/
$ gitdist dist-repo-status
```

(where `gitdist` is the updated script in your path)

That should skip over the Trilinos/ dir and move down to the charonMetaRepo/ dir and run gitdist from there and print the status table starting with charonMetaRepo/.

If setting and env var is an acceptable solution, then I can add documentation and automated testing and push to the 'master' branch.  That way, you would not need to create your own script for this behavior.

Let me know if you are interested.

bartlettroscoe:
> The reason I checked for the existence of a `.git` directory as opposed to the `.gitdist` or `.gitdist.default` files is it looked like it was possible to run gitdist without either of those files—if it didn't find them, it would assume their contents to be `..`

If the `gitdist` script can't find the `.gitdist` or `.gitdist.default` files, then it should just run in the current directory.  If one of those two files have not been set, then gitdist is pretty worthless.

Again, think of the use case where a user will use git to manage their home directory files (see, for example, http://www.bitflop.dk/tutorials/keeping-your-home-in-git.html).  If they do that, then looking for the base most `.git/` directory will result in gitdist always running in the user's home directory.  That would make gitdist worthless on that system.

Again, let me know if you are okay with having to set `export GITDIST_MOVE_TO_BASE_DIR=EXTREME_BASE` to get the behavior you want.

jmgate:
@bartlettroscoe, this all sounds good to me.  I also just pushed 442848e, which I think gives you the `moveToBaseDirMode = moveToImmediateBase` functionality you'd like.  It works for my mocked up setup, but you should try it on your home directory to see if it actually works.  I'd never heard of managing your home directory as a git repo&mdash;thanks for the link.

bartlettroscoe:
@jmgate, this looks great.  Now I will need to add some unit testing and documentation before this gets merged to the 'master' branch.  I will likely need to refactor this some to make it easier to unit test as well, but we will see.  But I many not be able to get to that cleanup until later next week (unless you need this merged urgently).

I think it is best to keep the default behavior not to move to the base directory since I think that should be the least surprising to people. But I suspect that once they read the documentation on this feature most users will want to set either `EXTREME_BASE` or `IMMEDIATE_BASE` (I will set `IMMEDIATE_BASE`).

Thanks for this great addition to the code!

jmgate:
Thanks, @bartlettroscoe&mdash;no rush on my end.  It'd be nice if it was available to users whenever I try to transition Charon, Drekar, and EMPIRE over to the meta-repo way of doing things, but that's still a ways out.

bartlettroscoe:
@jmgate, this is now merged and you should be able to use this on the 'master' branch of TriBITS.

I have created #212 to follow up with adding the automated tests and documentation needed to complete this.

jmgate:
Sounds good.  Thanks @bartlettroscoe.

PR #205, opened 2017-07-12
Colorize git commands run through gitdist

jmgate:
It was bugging me that `gitdist` stripped out whatever colors would normally be displayed when running the equivalent `git` commands from the command line.  This is an attempt to fix that problem by tying the `stdout` and `stderr` of the `subprocess` used to execute the `git` command to the `stdout` and `stderr` of the process running `gitdist` itself.  Not sure if that's the best way to do this&mdash;I only have about an hour's worth of experience with Python's `subprocess`.  I haven't seen any unintended side effects of this change yet, though I am by no means a power `gitdist` user.  Let me know if this just isn't a good idea, or if we need to go about it a different way.  Many thanks.

### DIFF
--- a/tribits/python_utils/gitdist.py
+++ b/tribits/python_utils/gitdist.py
@@ -786,6 +786,7 @@ def runCmnd(options, cmnd):
     print(cmnd)
   else:
     subprocess.Popen(cmnd, stdout=sys.stdout, stderr=sys.stderr).communicate()
+    print("")
 
 
 # Determine if a command exists:
--- a/tribits/python_utils/gitdist.py
+++ b/tribits/python_utils/gitdist.py
@@ -785,11 +785,7 @@ def runCmnd(options, cmnd):
   if options.noOpt:
     print(cmnd)
   else:
-    child = subprocess.Popen(cmnd, stdout=subprocess.PIPE).stdout
-    output = child.read()
-    sys.stdout.flush()
-    print(s(output))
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
This is so odd, but I never got an email from GitHub about the creation of this issue!  Is GitHub broken or something?  Is the SNL system filtering these emails as spam?  What is going on?

jmgate:
> Is GitHub broken or something?  Is the SNL system filtering these emails as spam?

@bartlettroscoe, I'm not sure.  I was messing with my notification settings last week to try to decrease the number of emails I get, and I think at this point I don't get any notifications from Trilinos or TriBITS, so I'm probably not the best person to ask.

bartlettroscoe:
The commit 0518ce8 I just pushed fixes the gitdist test suite with Python 2.6.6.

I will go ahead and merge and push this to the 'master' branch.

Thanks for fixing this!  It has annoyed me for a long time that running gitdist removed color (but color still gets removed when I pipe it into `less` for some reason).

jmgate:
> (but color still gets removed when I pipe it into `less` for some reason)

That's because command output gets buffered as it's sent through the pipe.  Command line fixes for that include prepending your command with `unbuffer` from the `expect` package or `stdbuf -o0 -e0` from, I think, GNU `coreutils`.  Unfortunately those fixes aren't necessarily portable.

Issue #200, opened 2017-07-05
Add some error checking and reporting for common TriBITS project developer and user mistakes


bartlettroscoe:
**Next Action Status:**

Added checks for usage of TRIBITS_PACKAGE(), TRIBITS_PACKAGE_POSTPROCESS(), etc. and some other things.  Next: Need to add checks for TRIBITS_SUBPACKAGE(), TRIBITS_SUBPACKAGE_POSTPROCESS(), etc.

**CC:** @fryeguy52 

**Description:**

The TriBITS framework can reduce some of the boiler plate raw CMake code that project developers need to write when setting up a CMake build and tests with CTest in their CMakeLists.txt files.  However, history has shown that project developers and TriBITS project users made some common mistakes when using TriBITS.  TriBITS already contains error checks and reporting for many types of common mistakes (e.g. misspelling the name of a TriBITS SE package, or trying to create a circular dependency) but there are many other types of mistakes that are not currently caught and reported very gracefully (e.g. not calling TRIBITS_PROJECT_POSTPROCESS() at the end of a packages's CMakeLists.txt file)

This Story is to add error checking and reporting (using `MESSAGE(FATAL_ERROR ...)`) for many common mistakes.  An initial list of the mistakes that can be caught and gracefully reported are:

* Check that `TRIBITS_PACKAGE()` is called before `TRIBITS_ADD_LIBRARY()` or `TRIBITS_ADD_EXECUTABLE()`. **[DONE]**

* Check that `TRIBITS_PACKAGE_DECL()`, `TRIBITS_PACKAGE_DEF()`, `TRIBITS_PACKAGE_POSTPROCESS()`, `TRIBITS_PROJECT()` and many other such user-facing macros and functions are called only once (i.e. per package for package-level functions and once per project for project-level functions). **[DONE]**

* Check that `TRIBITS_PACKAGE_POSTPROCESS()` is called before the end of the package's CMakeLists.txt file after all other commands like `TRIBITS_ADD_LIBRARY()` and `TRIBITS_ADD_EXECUTABLE()`. **[DONE]**

* Check that all of the arguments to `TRIBITS_PROJECT_DEFINE_DEPENDENCIES()` match named arguments (e.g. look for non-empty `PARSE_UNPARSED_ARGUMENTS` after calling `CMAKE_PARSE_ARGUMENTS(PREFIX ...)` and abort if found). **[DONE]**

[...]

PR #232, opened 2017-10-19
Subpkg user errors 200

fryeguy52:
More checks for catching user errors related to usage of subpackages (see #200).


bartlettroscoe:
@fryeguy52,

All the changes and the tests look reasonable.  I will look more closely before I start squashing commits and then doing the final test and push.

There are a few more checks that would be good to add that I thought of when looking over this:

* Check that TRIBITS_PACKAGE() is not called in the base CMakeLists.txt file for a package that has subpackages.  Instead, they must call TRIBITS_PACKAGE_DECL() and TRIBITS_PACKAGE_DEF() (with TRIBITS_PROCESS_SUBPACKAGES() called in between).

* Check that missing a call to TRIBITS_PROCESS_SUBPACKAGES() is caught if one forgets to call it before TRIBITS_PACKAGE_DEF() is called.

* Check that if only TRIBITS_PACKAGE_DECL() is called in parent packages's CMakeList.txt file that TriBITS reports the user needs to call TRIBITS_PROCESS_SUBPACKAGES() next.

* Check that TRIBITS_PACKAGE_DEF() is called once and only once in a parent packages's CMakeLists.txt file.

* Check that a subpackages's base CMakeLists.txt file does not call any of the commands TRIBITS_PACKAGE(), TRIBITS_PACKAGE_DECL(), TRIBITS_PACKAGE_DEF(), TRIBIITS_PROCESS_SUBPACKAGES(), TRIBITS_PACKAGE_POSTPROCESS().

* Check that TRIBITS_ADD_TEST_DIRECTORIES() and TRIBITS_ADD_EXAMPLE_DIRECTORIES() is called between TRIBITS_SUBPROJECT() and TRIBITS_SUBPROJECT_POSTPROCESS() (and the same for a top-level package's commands).

I also though of some other dumb things that users could do that we should test for:

* Calling any of the commands TRIBITS_PACKAGE(), TRIBITS_PACKAGE_DECL(), TRIBITS_PACKAGE_DEF(), TRIBITS_PROCESS_SUBPACKAGES(), TRIBITS_PACKAGE_POSTPROCESS() are not called in a package's base CMakeLists.txt file. That is, we should not set CACHE vars but instead use DUAL_SCOPE_SET() to set some of these vars.

* Calling any of the commands TRIBITS_SUBPACKAGE(), and TRIBITS_SUBPACKAGE_POSTPROCESS() are not called in a subpackage's base CMakeLists.txt file. That is, we should not set CACHE vars but instead use DUAL_SCOPE_SET() to set some of these vars.

Can you think of any other crazy mistakes that users might make with these commands?

bartlettroscoe:
I added the above list of checks to the list in #200.

bartlettroscoe:
@fryeguy52, are all of the commits in and are ready to review, clean up/merge commits and push?

fryeguy52:
All that remains is:

 "Check that if only TRIBITS_PACKAGE_DECL() is called in parent packages's CMakeList.txt file that TriBITS reports the user needs to call TRIBITS_PROCESS_SUBPACKAGES() next."

Should this be a warning?

bartlettroscoe:
> "Check that if only TRIBITS_PACKAGE_DECL() is called in parent packages's CMakeList.txt file that TriBITS reports the user needs to call TRIBITS_PROCESS_SUBPACKAGES() next."
> 
> Should this be a warning?

This may be poorly worded but we are looking for a hard error here.  Let's talk tomorrow when we meet.

fryeguy52:
I implemented all that we talked about last week this is ready for review

bartlettroscoe:
I am testing with Trilinos and I am finding only a few examples where there is bad usage.  I am not seeing the problem with STK that was reported in trilinos/Trilinos#218.  Turns out it was fixed 5 months ago.

I will now look over the code and tests more carefully and then squash and clean up commits.

bartlettroscoe:
I squashed and pushed the commit.  Thanks @fryeguy52!

PR #233, opened 2017-10-20
Kokkos config 207

kruger:
More testing is likely needed, but sufficient to have a PR to start the review.


kruger:
I just need to know which platform you want me to hop onto and test.
Sample config files would be useful.

PR #237, opened 2017-11-02
GITDIST_MOVE_TO_BASE_DIR Documentation and Testing

### LINKED ISSUE #212: Add support for runnning gitdist from subdir but move to base dir
**CC:** @jmgate 

**Description:**

Currently, the gitdist script runs in the current working directory.  If there is a .gitdist or .gitdist.default file in that directory, then it will loop over all of the existing git repos listed there.  That requires constantly cd'ing to the base git repo to run gitdist.  It would be nice if gitdist could be configured to allow it to be run in a subdir or sub git repo and move to the base repo where the .gitdist and .gitidst.default file exists and run from there.

ToDo: More fully document the use cases.

**Definition of Done:**

* Implements desired behavior **[DONE]**
* New behavior is defined and protected with automated tests.
* New behavior is documented with `gitdist --dist-help=move-to-base-dir`.

jmgate:
This PR is to contain the documentation and unit testing of the `GITDIST_MOVE_TO_BASE_DIR` feature we created in #203.

Closes #212.

### DIFF
--- a/test/python_utils/gitdist_UnitTests.py
+++ b/test/python_utils/gitdist_UnitTests.py
@@ -1536,7 +1536,6 @@ class test_gitdist(unittest.TestCase):
       os.environ["GITDIST_MOVE_TO_BASE_DIR"] = "EXTREME_BASE"
       cmndOut = GeneralScriptSupport.getCmndOutput(gitdistPathMock+" status")
       cmndOut_expected = \
-        "\n*** Base Git Repo: MockProjectDir\n" \
         "\n*** Base Git Repo: somewhere\n" \
         "['mockgit', 'status']\n\n"
       self.assertEqual(s(cmndOut), s(cmndOut_expected))
--- a/test/python_utils/gitdist_UnitTests.py
+++ b/test/python_utils/gitdist_UnitTests.py
@@ -103,7 +103,7 @@ class test_createAsciiTable(unittest.TestCase):
 
 
   def setUp(self):
-    self.gitdistMoveToBaseDir = os.environ["GITDIST_MOVE_TO_BASE_DIR"]
+    self.gitdistMoveToBaseDir = os.environ.get("GITDIST_MOVE_TO_BASE_DIR", "")
     os.environ["GITDIST_MOVE_TO_BASE_DIR"] = ""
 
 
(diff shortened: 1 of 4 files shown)


jmgate:
So far this just has changes to the unit test suite such that all tests pass if you have `GITDIST_MOVE_TO_BASE_DIR` set to something before running the test suite.  Working on unit testing now&hellip;

bartlettroscoe:
@jmgate, thanks!  There was one use case in particular that caused gitdist to hang when you you set the env var.  

However, I don't see any documentation added for this.   Should there be a `--dist-help=move-to-base-dir` output option added for this?  I don't see any modifications to the `gitdist.py` script itself.

jmgate:
Yup, I'm working on it right now.  Pushing soon&hellip;

jmgate:
Tests in place&mdash;adding documentation now&hellip;

jmgate:
Pretty sure this PR is good to go now as well.

bartlettroscoe:
@jmgate,

Thanks for adding all of this.  I think this is a nice feature to add to the script.

I added some commits to your topic branch to change the formatting of the documentation a little and update the TriBITS Developers Guide for this.

I also added a commit with a unit test for a test case that causes gitdist to hang that I have noticed.  That is, if you set GITDIST_MOVE_TO_BASE_DIR=IMMEDIATE_BASE or EXTREME_BASE and then run gitdist in a git repo where there is no .gitdist[.default] file in the parent git repos, then it just hangs.  The documentation says it should just run gitdist in the current working directly.

Can you take a look at this use case and see if you can fix the code so that it has the expected behavior?

jmgate:
Hmmm, I thought I had that case handled.  Yeah I'll look into it today.

jmgate:
Ah, looks like it works as documented for the default or EXTREME_BASE, but not for IMMEDIATE_BASE.  Let me check the implementation real quick.

jmgate:
@bartlettroscoe, I updated the implementation for IMMEDIATE_BASE and now I'm seeing the expected behavior for all cases.  Can you confirm?

jmgate:
Also added a unit test for when GITDIST_MOVE_TO_BASE_DIR is set to something invalid.  I think these changes are good to go again.

bartlettroscoe:
Looks great.  I am merging and pushing now with checkin-test.py script for TriBITS.

`GITDIST_MOVE_TO_BASE_DIR=IMMEDIATE_BASE` will be my new default.

PR #239, opened 2017-11-08
Add generic override for the c standard

prwolfe:
This modifies the TRIBITS_DEFINE_STANDARD_COMPILE_FLAGS_VARS
macro to respect a ${PROJECT_NAME}_C_Standard value when passed.


prwolfe:
@bartlettroscoe, 

I think I might should have sent this to another repo than TriBITSPub - let me know if I forked the wrong one.

Thanks!

Paul

bartlettroscoe:
@prwolfe, this is the right repo.  I will look at a way to write a test for this and add documentation then I will push to your branch.

PR #248, opened 2018-01-25
Add a workaround to prevent test script from being run twice

KyleFromKitware:
This also removes the need to set CTEST_COMMAND and friends, since it is no longer attempting to run it.

### DIFF
--- a/tribits/ctest_driver/TribitsCTestDriverCore.cmake
+++ b/tribits/ctest_driver/TribitsCTestDriverCore.cmake
@@ -1659,18 +1659,13 @@ FUNCTION(TRIBITS_CTEST_DRIVER)
   ENDIF()
 
   #
-  # Write a few variables to the global level to make cmake happy.
-  #
-  # If these don't get set in the base CTest script scope, CTest returns an
-  # error!
+  # This is a workaround to prevent the test script from being run twice. Since
+  # we are declaring tests inside a function, CTEST_RUN_CURRENT_SCRIPT gets set
+  # in the function scope, NOT the global scope. Manually setting it here with
+  # PARENT_SCOPE sets it at the global scope.
   #
 
-  SET(CTEST_SOURCE_DIRECTORY ${CTEST_SOURCE_DIRECTORY} CACHE INTERNAL "")
-  SET(CTEST_BINARY_DIRECTORY ${CTEST_BINARY_DIRECTORY} CACHE INTERNAL "")
-  IF ("${CTEST_COMMAND}" STREQUAL "")
-    SET(CTEST_COMMAND ctest)
-  ENDIF()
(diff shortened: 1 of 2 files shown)


bartlettroscoe:
This generated some failures when I ran it locally and submitted to CDash just now as shown at:

* https://my.cdash.org/index.php?project=TriBITS&parentid=1356987

It looks like the updated code generates the failure:

```

CMake Error: Some required settings in the configuration file were missing:
CTEST_SOURCE_DIRECTORY = /ascldap/users/rabartl/Trilinos.base/Trilinos/TriBITS/tribits/examples/TribitsExampleProject
CTEST_BINARY_DIRECTORY = /ascldap/users/rabartl/Trilinos.base/BUILDS/GCC-4.8.3/TRIBITS_MPI_DEBUG/BUILD
CTEST_COMMAND = (Null)
```

I will try putting the code:

```
-  # Write a few variables to the global level to make cmake happy.
-  #
-  # If these don't get set in the base CTest script scope, CTest returns an
-  # error!

-  SET(CTEST_SOURCE_DIRECTORY ${CTEST_SOURCE_DIRECTORY} CACHE INTERNAL "")
-  SET(CTEST_BINARY_DIRECTORY ${CTEST_BINARY_DIRECTORY} CACHE INTERNAL "")
-  IF ("${CTEST_COMMAND}" STREQUAL "")
-    SET(CTEST_COMMAND ctest)
-  ENDIF()
-  SET(CTEST_COMMAND ${CTEST_COMMAND} CACHE INTERNAL "")
```

and see if that allows the tests to pass.

P.S. I need to figure out how to set up tests on Travis  CI with different versions of CMake so that I can run all of these tests with different versions of CMake.

bartlettroscoe:
After that commit https://github.com/TriBITSPub/TriBITS/pull/248/commits/d9bb94fcd4862181df045e4b492c1d7e8999445b, all tests passed by one:

* https://my.cdash.org/index.php?project=TriBITS&parentid=1356995

But that test failure is unrelated to this (seems that 'make VERBOSE=1' does not always yield a verbose make?).

I will go ahead and merge and push this.  I will then test this on the real application that is causing trouble on 'shiller' ...

KyleFromKitware:
Don't merge yet. The problem is that I set with PARENT_SCOPE and not INTERNAL CACHE. The error you're seeing with CTEST_COMMAND missing shouldn't be appearing at all if my fix is working. I'm working on an updated version now.

bartlettroscoe:
> Don't merge yet. The problem is that I set with PARENT_SCOPE and not INTERNAL CACHE. The error you're seeing with CTEST_COMMAND missing shouldn't be appearing at all if my fix is working. I'm working on an updated version now.

I already merged but I can revert the merge and accept your updated branch once you have it ready.

Thanks!

PR #249, opened 2018-01-25
Manually set CTEST_RUN_CURRENT_SCRIPT at root

KyleFromKitware:
Setting CTEST_RUN_CURRENT_SCRIPT with PARENT_SCOPE didn't work because there was an instance where ctest_start() was nested inside two functions. I have instead set it at the root of each test script, which is consistent with what CMake does in its internal test scripts.

I have also removed CTEST_COMMAND and friends because setting them should never be necessary if everything is working properly. If any of the tests give any errors about CTEST_COMMAND not being found, that means I missed one. Please let me know and I will fix it.

Please revert my previous PR and use this one instead.


bartlettroscoe:
@KyleFromKitware,

Thanks for the update.  The only problem with this solution is that it is not scalable.  This fixes the TriBITS driver scripts but it does not address the Trilinos, CASL VERA, MPACT, COBRA-TF, and every other project's CTest TriBITS driver scripts.

Is there a way to just abort the ctest -S program and control what it returns?

KyleFromKitware:
We could simply set `CTEST_COMMAND` to a no-op like `echo` or `cmake -E echo`. I will try it and see if it works.

I also have a proper fix for this bug in the test pipeline for mainline CMake. Once it's released, this hack won't be necessary anymore with the new version.

bartlettroscoe:
> We could simply set CTEST_COMMAND to a no-op like echo or cmake -E echo. I will try it and see if it works.

Thanks.  Let me know if that works.

PR #250, opened 2018-01-26
Set CTEST_COMMAND to a no-op

KyleFromKitware:
As discussed, I have set `CTEST_COMMAND` to a no-op, since this seems to be the most scalable solution to this problem.


bartlettroscoe:
Excellent!  This seems to have been the fix.   Ran the fuller TriBITS test suite and submitted to CDash as shown at:

* https://my.cdash.org/index.php?project=TriBITS&parentid=1357632

and I tried this with Trilinos on the build on `shiller` where I was seeing this problem and the problem disappeared.

I will clean up the PR commits (after backing out my merge commit) and then test again and push to TirBITS master and then update the snapshot of TriBITS in Trilinos to fix this problem for Trilinos.

Issue #272, opened 2019-01-17
Error Checking in Package Macros


kevindugan:
In certain packages of VERA, the following logic is followed:

```
TRIBITS_PACKAGE_DECL(PackageName)
 
TRIBITS_PROCESS_SUBPACKAGES()
 
[other Tribits macro]
 
TRIBITS_PACKAGE_DEF()
```

where `[other Tribits macro]` can be `TRIBITS_ADD_TEST_DIRECTORIES` or `TRIBITS_ADD_EXECUTABLE` for example.

I think that the error checking should be based on the variable `${PACKAGE_NAME}_TRIBITS_PACKAGE_DECL_CALLED` instead of `${PACKAGE_NAME}_TRIBITS_PACKAGE_CALLED` in these cases to allow for the forward declaration of a package.

PR #273, opened 2019-01-17
Modifying error checking in Package Macros

kevindugan:
This PR addresses errors generated when using `TRIBITS_PACKAGE_DECL` instead of `TRIBITS_PACKAGE` (see #272),


bartlettroscoe:
@kevindugan, looks like this broke the TriBITS test suite.  See:

* https://testing-vm.sandia.gov/cdash/testDetails.php?test=64459333&build=4424023

kevindugan:
@bartlettroscoe I think the CI failed because I changed the error message to include `PACKAGE_DECL`. I'll need to update the test as long as you're ok with the error message reading

```
Must call TRIBITS_PACKAGE() or TRIBITS_PACKAGE_DECL() before ...
```
instead of

```
Must call TRIBITS_PACKAGE() before ...
```

bartlettroscoe:
We will also need to an an automated test that triggers the defect before the change and verifies that the change to the source code fixes the defect.  I will see if I can add that later tonight.

-Ross

bartlettroscoe:
@kevindugan, does this just change the error message or does it fix the behavior reported in #272?

bartlettroscoe:
@kevindugan, what I mean is, does applying this PR actually fix the usage of TriBITS for the CASL VERA case you are dealing with here or does more need to be done?

kevindugan:
@bartlettroscoe As far as I can tell now, these changes don't fix the underlying problem in VERA. These changes only allow me to upgrade Tribits to the most recent version. I think there is still something else that we need to resolve.

bartlettroscoe:
> @bartlettroscoe As far as I can tell now, these changes don't fix the underlying problem in VERA. These changes only allow me to upgrade Tribits to the most recent version. I think there is still something else that we need to resolve.

@kevindugan, that is what I thought.  I will go ahead and merge this PR.  But we need to reproduce the failing use case in a native TriBITS test so that we can fix it.  Let's discuss that it #272.

kevindugan:
Hi ross, I'm currently working to add a test which is resolved by this PR. Could we wait to merge until we have that test added?

bartlettroscoe:
> Hi ross, I'm currently working to add a test which is resolved by this PR. Could we wait to merge until we have that test added?

Shoot, I already merged.  We can create a new PR from your same branch.

kevindugan:
> Shoot, I already merged. We can create a new PR from your same branch.

Lol, ok. No worries. I'll open up a new PR with the test.

PR #275, opened 2019-01-22
Adding failing test for Issue #272

kevindugan:
This PR adds regression tests for Issue #272. These tests fail without the changes implemented in PR #273 and pass when those changes are applied.

### DIFF
--- a/test/core/ExamplesUnitTests/CMakeLists.txt
+++ b/test/core/ExamplesUnitTests/CMakeLists.txt
@@ -3360,6 +3360,105 @@ TRIBITS_ADD_ADVANCED_TEST( TribitsExampleProject_PkgWithUserErrors_PACKAGE_INIT
 
   )
 
+TRIBITS_ADD_ADVANCED_TEST( TribitsExampleProject_PkgWithUserErrors_PACKAGE_DECL
+  OVERALL_WORKING_DIRECTORY TEST_NAME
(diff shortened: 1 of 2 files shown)


kevindugan:
@bartlettroscoe I placed these tests in the User Error Package, but would it be better to create a new 'PkgRegressions' package? I didn't want to change the structure too much for these tests, but I'll admit it seems a little strange to put valid package declarations in something testing for user errors.

bartlettroscoe:
@kevindugan, does this PR fix the usage of TriBITS by VERA?

kevindugan:
@bartlettroscoe This PR only addresses the testing of the changes I made in the previous PR (#273). I still need to find a solution to the VERA/MOOSE problem I'm having. 

On a related issue, we're seeing debug/optimized tags leaking into the libraries we're linking to. I've found where these tags are added for certain libraries (HDF5, I think). Is it possible that these tags aren't getting removed from another library, Qt maybe?

bartlettroscoe:
> This PR only addresses the testing of the changes I made in the previous PR (#273). I still need to find a solution to the VERA/MOOSE problem I'm having.

@kevindugan, okay, then lets address the real problem in getting VERA to work.  In issue #272, can you please attach the full CMakeLists.txt file where this error is triggered and then attach the full configure output?  Otherwise, I will need to get back my access to ORNL so that I can see what is happening myself.  Let's discuss in issue #272.

Issue #276, opened 2019-02-07
Add RawAndTribitsHelloWorld example project and howto


bartlettroscoe:
We need a TriBITS example and howto for putting a TriBITS build system and a raw CMake build system side-by-side.  This was requested by a Trilinos developer but this is something that everyone should know how to do.

bartlettroscoe:
@csiefer2,

The long-promised and howto and example for putting a raw CMake build system and a TriBITS build system side-by-side is pushed to TriBITS shown at:

* https://tribits.org/doc/TribitsDevelopersGuide.html#how-to-put-a-tribits-and-raw-cmake-build-system-side-by-side

and:

* https://github.com/TriBITSPub/TriBITS/tree/master/tribits/examples/RawAndTribitsHelloWorld

If you clone the TriBITS repo, you can test this out yourself as described the above howto and in the README:

* https://github.com/TriBITSPub/TriBITS/blob/master/tribits/examples/RawAndTribitsHelloWorld/README

Please let me know if this is not clear and I will try to improve it to address your questions.

Sorry for the very long delay in getting this done.

bartlettroscoe:
@csiefer2,

Ever have a chance to look at this?  Still looking for a review before I close this issue.

csiefer2:
I can now make cmake segfault :)

bartlettroscoe:
@csiefer2 said:

> Evidently my tribits_add_library() call inside my child directory's CmakeLists.tribits.cmake causes cmake to woof.

Any chance you accidentally have a recursive file include or something?  That can cause CMake to crash.

csiefer2:
I don't think so.

bartlettroscoe:
@csiefer2, can we close this?  Are you good to go with this?

bartlettroscoe:
@csiefer2 reported to me in person a few months ago that this looked good for him.

Closing as complete.

PR #277, opened 2019-02-13
Add RawAndTribitsHelloWorld example (#276)

bartlettroscoe:
This example shows how to put a raw CMake build system and a TriBITS build
system side-by-side.  (See #276).

There is also a "how to" based on this example.  Hopefully this will be enough
to explain how to accomplish this.


bartlettroscoe:
I used checkin-test.py script to push which amended the commit so GitHub does not realize that this has been pushed.

Closing as complete.

PR #283, opened 2019-03-28
Add support for version date info

bartlettroscoe:
This is the basic TriBITS functionality to address trilinos/Trilinos#4532.  But doing it in TriBITS we can add very strong automated testing to protect this.


bartlettroscoe:
@fryeguy52,

This has already been merged, but can you look this over some and especially the documentation at:

* https://tribits.org/doc/TribitsBuildReference.html#generating-git-version-date-files

and the generated header file from [Tribits_version_headers.h.in](https://github.com/TriBITSPub/TriBITS/pull/283/files#diff-8577c552306eefc5cef692698ab0d0b3)

Users will be looking at that documentation so it would be good to get this reviewed (if nothing else).

bartlettroscoe:
Done a long time ago and has been used by SPARC for many months.

PR #291, opened 2019-06-12
Changed to use print-function for python 3 compatability.

prwolfe:
These two files were failing in testing for python 3
and the only issue was use of the newer print. This
works under both python 2 and 3.


bartlettroscoe:
@prwolfe, thanks for fixing this!  Need to figure out how to update the .travis.yaml file to also test with Python3 as well.

prwolfe:
No problem - any idea when this migrates into Trilinos? I am going to add python 2 and 3 testing to the PR system after if does.

bartlettroscoe:
@prwolfe asked:

> any idea when this migrates into Trilinos?

Following the instructions at:

* http://trac.trilinos.org/wiki/TriBITSTrilinosDev#update_snapshot

I will submit a Trilinos PR right now.

bartlettroscoe:
@prwolfe, the PR is https://github.com/trilinos/Trilinos/pull/5372.  Can you please approve that so it can be merged?

PR #296, opened 2019-10-07
Added howto: globally changing X_ENABLE_Y during package config

brian-kelley:
Following discussion in Trilinos issue 6041, where this was used.

Link to discussion: https://github.com/trilinos/Trilinos/issues/6041

### DIFF
--- a/tribits/doc/developers_guide/TribitsDevelopersGuide.rst
+++ b/tribits/doc/developers_guide/TribitsDevelopersGuide.rst
@@ -3665,12 +3665,18 @@ In more detail, these rules/behaviors are:
 
 22) **TriBITS auto-enables/disables done using non-cache local variables**:
     TriBITS setting (or overrides) of enable/disable cache variables are done
-    by setting local non-cache variables at the top project-level scope.  This
-    is done so they don't get set in the cache and so that the same dependency
+    by setting local non-cache variables at the top project-level scope
+    (i.e. the ``<projectDir>/CMakeLists.txt`` file scope).  This is done so
+    they don't get set in the cache and so that the same dependency
     enable/disable logic is redone, from scratch, with each re-configure.
     This results in the same enable/disable logic output as for the initial
     configure.  This is to avoid confusion by the user about why some SE
     packages and TPLs are enabled and some are not on subsequent reconfigures.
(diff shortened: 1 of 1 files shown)


tribits/doc/developers_guide/TribitsDevelopersGuide.rst
```diff
@@ -6038,6 +6038,25 @@ Enable/Disable Logic`_).  Also, these files get processed in `Reduced Package
...
+Changing the value of ``<XXX>_ENABLE_<YYY>`` inside package `<XXX>` using ``SET``
+changes the variable's value inside the package's scope, but all other
+packages will see the old value of ``<XXX>_ENABLE_<YYY>``.
+
+However, it is sometimes necessary to make this change visible to downstream packages.
+To do this, use ``DUAL_SCOPE_SET(<XXX>_ENABLE_<YYY> <val>)`` instead. This
+sets the value in both the global (project) scope and in the local scope of package `<XXX>`.
+Any downstream package (configured after `<XXX>` chronologically)
+will see the new value: ``${<XXX>_ENABLE_<YYY>} STREQUAL <val>``.
+Caution: packages that don't depend on `<XXX>` may or may not see the change.
```
bartlettroscoe: @brian-kelley, what is the purpose of:

> Caution: packages that don't depend on `<XXX>` may or may not see the change.

?

Under what situations would an upstream package change some variable `<XXX>_ENABLE_<YYY>` and a downstream package's `<packageDir>/CMakeLists.txt` file would not see it?  If if it did not see it, why do we care (because one would assume it was not dependent on that variable).

brian-kelley: @bartlettroscoe What I meant by this is that the change won't be visible to any non-downstream package. But I agree that a) we don't care and b) this is kind of reiterating the last sentence in a more confusing way.

bartlettroscoe:
@brian-kelley, can you please have a look at the updated text?  I tried to provide some more context for people that may not know the issues as well and provide some links to relevant material.

If this looks good to you, I can merge.  Otherwise, please make whatever changes you think would be good.

brian-kelley:
@bartlettroscoe Yes, that looks good.

PR #298, opened 2020-01-02
Silence CMake warning when EXTRAREPOS_FILES is empty

sethrj:
Our repository has no "extra_repos" file, so CMake emits a warning during every configure, presumably because "" is not a valid filename.

```
CMake Warning (dev) at TriBITS/tribits/core/utils/AdvancedSet.cmake:57 (SET):
  implicitly converting 'FILENAME' to 'STRING' type.
Call Stack (most recent call first):
  TriBITS/tribits/core/package_arch/TribitsGlobalMacros.cmake:751 (ADVANCED_SET)
  TriBITS/tribits/core/package_arch/TribitsProjectImpl.cmake:128 (TRIBITS_DEFINE_GLOBAL_OPTIONS_AND_DEFINE_EXTRA_REPOS)
  TriBITS/tribits/core/package_arch/TribitsProject.cmake:93 (TRIBITS_PROJECT_IMPL)
  CMakeLists.txt:174 (tribits_project)
This warning is for project developers.  Use -Wno-dev to suppress it.
```


bartlettroscoe:
Interesting.   This technically changes the interpretation of this cache var but I think it should be okay.  Let's see what the Travis CI TriBITS test suite says.

sethrj:
I realize now from your last email that it's unusual to *not* have an EXTRAREPOS_FILE. That probably explains why you (and most projects) haven't seen this before.

bartlettroscoe:
The Travis CI checks passed so I think it is safe to merge this.

The only use case that I know of that would break with this change is if someone set:

```
-D<Project>_EXTRAREPOS_FILES=<some-relative-path>/MyExtraRepositoriesList.cmake
```

and expected `<some-relative-path>/MyExtraRepositoriesList.cmake` to be relative to the build directory instead of the source tree.  But I don't think anyone would actually do that so I don't think this is going to break any existing customer's code.

Therefore I will go ahead and merge this.

@sethrj, thanks for the PR!

Issue #299, opened 2020-02-05
Move to modern CMake target-based propagation of build information


bartlettroscoe:
**Parent Issue:**
* #367

**Child Issues:**
* #433
* #443

**Blocked By:** trilinos/Trilinos#8498

## Description

One of the things holding back TriBITS and projects using TriBITS in accessing new features of CMake for which TriBITS still using variables to manage compiler options and include directories.  This issue is to move to using:

* target_compile_definitions()
* target_compile_features()
* target_compile_options()
* target_include_directories()
* target_link_directories()
* target_link_libraries()
* target_link_options()

and to move completely to targets instead of variables for each TriBITS (internal and external) package.

This needs to be done as part of a comprehensive refactoring as part of #63.

This refactoring should implement and respect [this proposed standard for `<Package>Config.cmake` files](https://docs.google.com/document/d/19gIZgPAfnK7RTdh0sMKr5Yeo1epPVKIfona-oxoCtsg).

Related to:
* SEPW-214

## Tasks

* [x] Add `PUBLIC` to `target_link_libraries()` in TriBITS intra-package linking
* [x] Add new library`${PACKAGE_NAME}::<lib-name>` targets and `${PACKAGE_NAME}::all_libs` INTERFACE library targets for TriBITS packages and `target_link_libraries( ... PUBLIC ... )` to define dependencies to upstream packages. (see PR #424)
* [x] Experiment with adding new `${TPL_NAME}::all_libs` INTERFACE GLOBAL library targets for TPLs and use `target_link_libraries( ... INTERFACE ... )` to define dependencies between TPLs.
* [x] Add include directories to all internal targets with `target_include_directories()`
* [x] Address problems with partial refactoring of TriBITS to modern CMake targets and get merge to Trilinos 'develop' to stick: #433 
* [x] Complete refactoring of TriBITS targets to use modern CMake targets for all dependencies and managing include directories (no more using variables `<Package>_INCLUDE_DIRS`, `<Package>_TPL_INCLUDE_DIRS`, `<Package>_LIBRARIES` or `<Package>_LIBRARY_DIRS`).
* [x] Strip out all old machinery in TriBITS involving vars `<Package>_INCLUDE_DIRS`, `<Package>_TPL_INCLUDE_DIRS`, `<Package>_LIBRARIES` and `<Package>_LIBRARY_DIRS`.
* [x] Do some testing with Trilinos, Albany, Nalu, and SPARC.
[...]

PR #300, opened 2020-02-11
cmake: Handle C++ standard selection and propagation via CMake features

bradking:
C++11 is now required by TriBITS clients.  Enable support unconditionally.  Use CMake's standard settings for specifying the C++ standard and propagating it to dependents.  Avoid hard-coding `-std=` flags, as doing so breaks CMake's model of the compiler's C++ standard modes.

Drop the  `${PROJECT_NAME}_ENABLE_CXX11` option and hard-code a variable of the same name to "ON" to preserve compatibility while clients are updated.


bradking:
See trilinos/Trilinos#6814 for the Trilinos updates using this change to TriBITS.

tribits/core/package_arch/TribitsGlobalMacros.cmake
```diff
@@ -1952,6 +1949,20 @@ MACRO(TRIBITS_SETUP_ENV)
     INCLUDE("${PROJECT_COMPILER_CONFIG_FILE}")
   ENDIF()
 
+  # Set up C++ language standard selection.
+  IF (NOT CMAKE_CXX_STANDARD)
+    SET(CMAKE_CXX_STANDARD 11)
```
bartlettroscoe: @bradking, just to be clear, but `-std=c++11` will no longer show up in `CMAKE_CXX_FLAGS` that gets exported in the generated `<Package>Config.cmake` or `<Project>Config.cmake` files, right?  I will be to test that for myself.

tribits/core/package_arch/TribitsLibraryMacros.cmake
```diff
@@ -864,6 +864,10 @@ FUNCTION(TRIBITS_ADD_LIBRARY LIBRARY_NAME_IN)
 
     TARGET_LINK_LIBRARIES(${LIBRARY_NAME}  ${LINK_LIBS})
 
+    IF (${PROJECT_NAME}_CXX_STANDARD_FEATURE)
+      TARGET_COMPILE_FEATURES(${LIBRARY_NAME} PUBLIC "${${PROJECT_NAME}_CXX_STANDARD_FEATURE}")
```
bartlettroscoe: Nice!  The first step toward modern CMake in #299.

tribits/ReleaseNotes.txt
```diff
@@ -2,7 +2,14 @@
 Release Notes for TriBITS
 ----------------------------------------
 
-2108/10/10:
+2020/02/07:
+
+(*) MAJOR: TriBITS Core: CMAKE_CXX_STANDARD is now always set to at least 11
```
bartlettroscoe: We need to also add that the `-std=c++11` option will be removed from `CMAKE_CXX_FLAGS` from the generated and installed `<Project>Config.cmake`, `<Package>Config.cmake`, `Makefile.export.<Project>`, and `Makefile.export.<Package>` files.  This does not really break CMake clients that pull in `<Project>Config.cmake` or `<Package>Config.cmake` because they will get what they need through the imported targets but it will break Makefile clients that expect to get `-std=c++11` through the `CMAKE_CXX_FLAGS` through the imported `Makefile.export.<Project>` or and `Makefile.export.<Package>` files.  But this does not need much of an explanation given the desire to drop support for these files as per #299.

tribits/core/package_arch/TribitsDefineStandardCompileVars.cmake
```diff
@@ -108,14 +108,8 @@ MACRO(TRIBITS_DEFINE_STANDARD_COMPILE_FLAGS_VARS  ENABLE_SHADOWING_WARNINGS)
   IF (${PROJECT_NAME}_CXX_STRONG_COMPILE_WARNING_FLAGS)
     SET(CXX_STRONG_COMPILE_WARNING_FLAGS ${${PROJECT_NAME}_CXX_STRONG_COMPILE_WARNING_FLAGS})
   ELSE()
-    IF (NOT ${PROJECT_NAME}_ENABLE_CXX11)
-       SET(CXX_98_OPTION " -std=c++98") # C++98 standard code
```
bartlettroscoe: Right, this was to enforce the C++98 standard for code we wanted to make sure did not allow to slip in C+11 features. No need for that anymore. In fact, this should have been removed a long time ago when Trilinos was made to require C++11.

tribits/core/package_arch/TribitsDefineStandardCompileVars.cmake
```diff
@@ -108,14 +108,8 @@ MACRO(TRIBITS_DEFINE_STANDARD_COMPILE_FLAGS_VARS  ENABLE_SHADOWING_WARNINGS)
...
-    ELSE()
-       SET(CXX_98_OPTION)
-    ENDIF()
     MULTILINE_SET(CXX_STRONG_COMPILE_WARNING_FLAGS
       ${COMMON_STRONG_COMPILE_WARNING_FLAGS}
-      "${CXX_98_OPTION}"
```
bartlettroscoe: Was dumb to add `-std=c++98` to warning flags. That was just bad programming on my part :-(

tribits/core/package_arch/TribitsGlobalMacros.cmake
```diff
@@ -254,12 +254,9 @@ MACRO(TRIBITS_DEFINE_GLOBAL_OPTIONS_AND_DEFINE_EXTRA_REPOS)
...
 
-  IF ("${${PROJECT_NAME}_ENABLE_CXX11_DEFAULT}" STREQUAL "")
-    SET(${PROJECT_NAME}_ENABLE_CXX11_DEFAULT OFF)
-  ENDIF()
-  ADVANCED_OPTION(${PROJECT_NAME}_ENABLE_CXX11
-    "Enable the C++11 compiler options and related code (see ${PROJECT_NAME}_CXX11_FLAGS)"
-    ${${PROJECT_NAME}_ENABLE_CXX11_DEFAULT} )
+  # Hard-code a variable with the same name as a now-removed option that is always enabled.
+  # This can be removed after clients have been updated.
+  SET(${PROJECT_NAME}_ENABLE_CXX11 ON)
```
bartlettroscoe: It would be nice to get rid of this at some point since this is going to be confusing for projects that support/require C++ standards newer than C++11, like C++14 (see trilinos/Trilinos#6260).  But since downstream projects may still have logic that looks for this, I guess it is a good idea for now to keep defining this.

It would be nice if CMake had a way to deprecate a CMake variable so that if the variable got read by client code then it would generate a deprecated warning.  Otherwise, I don't know how to get a client to stop using a CMake variable of a given name :-(

tribits/core/package_arch/TribitsCXX11Support.cmake
```diff
@@ -1,257 +0,0 @@
-# @HEADER
```
bartlettroscoe: So nice to rip out large amounts of code :-)

tribits/core/package_arch/TribitsGlobalMacros.cmake
bradking: Correct.  I'll update the release notes to mention that.

tribits/ReleaseNotes.txt
bradking: Thanks.  I've updated the release notes for this.

bradking:
> I will need to carefully test ... before I merge this but this looks great.

Good.  I've removed the `WIP:` prefix from the PR title because this can be merged without immediately affecting Trilinos.

PR #225, opened 2017-09-18
Modify install-cmake.py to build CMake with SSL (#220)

bkochuna:
For CMake to be able to submit to a drop site via HTTPS
the cmake/ctest executables must have been built with OpenSSL
support. This commit modifies the install-cmake.py script to:
  * Check if CMake is installed on the system
  * Configure CMake with CMake and OpenSSL support

The CMakeLists.txt for configuring CMake with CMake with is smart
enough to allow you to still build CMake even if you request OpenSSL
and it is not installed. The only requirement is that CMake be
configured with CMake instead of autotools. If CMake is not available
then install-cmake.py defaults to its previous behavior.

Issue #167, opened 2016-12-13
Create reasonable delivery mechanism for CMake+Ninja+Fortran


bartlettroscoe:
To fully take advantage of dummy makefiles for Ninja, we need to provide ninja and CMake 3.7.0+ to support Fortran.  This is needed by CASL VERA for example but I suspect will also be needed for many Sandia projects as well.

This can be accomplished with:

1) Add cmake-3.7.0 to install-cmake.py script

2) Add a install-ninja.py script and have it install ninja_fortran/1.7.2 (see https://github.com/Kitware/ninja/releases/tag/v1.7.2.gcc0ea.kitware.dyndep-1)

3) Add install-ninja.py to install-devtools.py (which CASL VERA uses)

These are already installed with some SEMS modules at Sandia (see https://sems.sandia.gov/confluence/pages/viewpage.action?pageId=1998951).

bartlettroscoe:
Now that people are using Spack and it is easy to manually install Ninja and CMake so I am going to clos this as "wontfix"

bartlettroscoe:
Spack is not a reasonable delivery system for simple tools like CMake and Ninja.  If you use Spack to try to build Ninja, it will first try to build and install Python, Pearl and several other packages.  Same goes for CMake.  That is insane.  We need a simple way to install these basic tools.

I saddens me to reopen this issue :-(

bartlettroscoe:
NOTE: As of [Ninja 1.10](https://github.com/ninja-build/ninja/releases/tag/v1.10.0), support for Fortran with CMake is now part of mainline Ninja.  Also, as of CMake 3.16, it will automatically support Fortran with Ninja 1.10+ (see https://gitlab.kitware.com/snl/project-1/-/issues/80).

Therefore, just installing raw Ninja 1.10+ will automatically provide what you need if you have CMake 3.16+.  And since we have the updated [TriBITS/tribits/devtools_install/install-cmake.py](https://tribits.org/doc/TribitsBuildReference.html#installing-cmake-from-source-developers-and-experienced-users) command, I think this is good enough.

Perhaps in the future I can provide a nice little `install-ninja.py` command like the `install-cmake.py` command to make this super easy.  But I will make a new issue for that later.

I am going to close this as WORKSFORME!

PR #306, opened 2020-03-11
Upgrade install-cmake.py to get arbitrary versions from GitHub and other improvements (#167)

bartlettroscoe:
This updates the install-cmake.py scirpt to allow downloading and installing arbitrary versions off of the GitHub site.  This is part of #167.

I also merged in the branch from PR #225 and made some adjustments to make it robust by default (which does not have OpenSSL support by default).


bartlettroscoe:
I am going to go ahead and merge this.  If we need to make tweaks, we can do that in a future PR.

PR #321, opened 2020-05-27
Add patch to log CTest environment variables

KyleFromKitware:


bartlettroscoe:
This was resolved with the upgrade to CMake 3.18 and there is no need for this patch.

PR #323, opened 2020-06-19
Remove warning when a component isn't found in an installed Config.cmake 

sethrj:
This isn't necessary as `find_package` will warn the first time optional components are missing and error out if they're required. It's intrusive if they're optional and not found, because this warning shows up at every reconfigure.


sethrj:
Thanks for the quick merge!

PR #332, opened 2020-09-23
Improve docuemntation for behavior of RUN_SERIAL logic (trilinos/Trilinos#7112)

bartlettroscoe:
This is an attempt to address review comments in PR trilinos/Trilinos#8063.


bartlettroscoe:
@e10harvey, this is to address any unresolved issues from https://github.com/trilinos/Trilinos/pull/8063#discussion_r493609955.  Let's please review TriBITS changes in TriBITS and not in Trilinos PRs (because the Trilinos PRs lack tests).

e10harvey (commented):
I hold that a warning should be printed on line 639 in `TRIBITS_SET_RUN_SERIAL` and that the override behavior should be clearly documented on line 636.

bartlettroscoe:
> I hold that a warning should be printed on line 639 in `TRIBITS_SET_RUN_SERIAL` and that the override behavior should be clearly documented on line 636.

@e10harvey, I think this is a case where we just disagree. From the user's perspective, if they set `<fullTestName>_SET_RUN_SERIAL=OFF` then they should have every expectation that that should will respected, regardless what is going on in the code.  With CMake, it should be that any option set in the cache by the user will override anything going on in the CMake code, if at all possible.  As a user, it is very frustrating to try to set a cache var and then have that ignored internally for some reason.  Now, if a cache var's value gets overwritten by some internal logic, then that **should** generate some type of printed comment/note to inform the user about this.  I guess this assumption should be written down somewhere if it not already but users really want this behavior because it allows them to not have to dig through CMake code to figure out why they cache variables they set are being ignored.

From the developer's perspective, they should be expected to read the updated documentation at:

https://github.com/TriBITSPub/TriBITS/pull/332/files#diff-533e6f3095d574099185299639e9dd91L158

so this behavior should not come as a surprise.

I guess we could ask some Trilinos developers what they think about producing a warning or not in this case.  But in my experience, my guess is that they will not care one way or the other (and most would be annoyed by a warning or would not even notice a warning if it was emitted, and it would just increase spam in the already full TriBITS configure output).

Perhaps TriBITS needs to introduce a log file for these types of NOTES and log them there instead of to STDOUT?  In fact, a lot of the text output currently going to STDOUT (like the mass of output for enable/disable logic for packages and TPLs) could go to a log file by default (unless the user asked that it goes to STDOUT).

e10harvey:
@bartlettroscoe: Fair enough. Thank you for explaining the reasoning behind the override behavior.

bartlettroscoe:
FYI: I had rebased the commit in this PR and forgot to push to this branch before merging locally and pushing as the commit 2850245d94de4dc12a7a3971876bb2e823d2681 as part of the merge commit 22f283084a037b0e5987aa19a7cf5bf376f17939.  After that I forced pushed the commit 2850245d94de4dc12a7a3971876bb2e823d2681 to this branch 'tril-7112-run-serial-3' and GitHub marked this PR as "Closed" instead of "Merged" which is kind of irritating.  But in any case this is PR is merged and closed.

PR #346, opened 2021-01-07
Fix wrapping of email lines exceeding 1000 characters

e10harvey:
RFC 821 states that the text line size including <CRLF> is 1000 characters. This text line size is causing spaces to be inserted into the email body and break long hyperlinks. With these changes, we use quoted-printable "Content-Transfer-Encoding" to ensure integrity of the data being sent through the gateway when html text lines are longer than 1000 characters.

Sorry for the changes that remove trailing whitespace; this was done automatically by my editor.

### DIFF
--- a/tribits/ci_support/CDashQueryAnalyzeReport.py
+++ b/tribits/ci_support/CDashQueryAnalyzeReport.py
@@ -78,7 +78,7 @@ def convertInputDateArgToYYYYMMDD(cdashProjectTestingDayStartTime, dateText,
 
 # Validate a date YYYY-MM-DD string and return a date object for the
 # 'datetime' module.
-# 
+#
 def validateAndConvertYYYYMMDD(dateText):
   try:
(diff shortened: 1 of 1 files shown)


bartlettroscoe (approved):
@e10harvey, thanks for the PR.  I never noticed a problem with this but I tested this locally on a COE RHEL7 machine with stock Python 2.7.5 with:

```
../TrilinosATDMStatus/trilinos_atdm_builds_status.sh --date=yesterday \
   --send-email-to=rabartl@sandia.gov
```

and the generated email in Outlook looked fine.  (There are no unit tests that check these lines of code so you must need to run and generate the email manually.)

I am sure all of the automated tests will pass when I push with the local  `checkin-test-crf450.sh` driver script (also because the unit tests never actually send out an email).

Thanks again!

bartlettroscoe:
Related to:
* SEPW-215

PR #347, opened 2021-01-11
Add expected builds filter

e10harvey:
TODOs:

- [x] Remove duplicate commits
- [x] Create option to filter by expected builds file

### DIFF
--- a/test/ci_support/CDashQueryAnalyzeReport_UnitTests.py
+++ b/test/ci_support/CDashQueryAnalyzeReport_UnitTests.py
@@ -1700,7 +1700,7 @@ class test_AddIssueTrackerInfoToTestDictFunctor(unittest.TestCase):
 
 #############################################################################
 #
-# Test CDashQueryAnalyzeReport.testsWithIssueTrackersMatchExpectedBuilds()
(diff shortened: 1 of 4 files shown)


bartlettroscoe (changes requested):
@e10harvey, thanks for the PR.  I can see what it is you want to do here.

I will put more details in a comment in:

* https://gitlab-ex.sandia.gov/atdm/TrilinosATDMStatus/-/merge_requests/4

but technically this code is not correct for the general use case (as described in the associated review comments below).

However, I know how to fix this and it will actually involve less new code (because 95% of the functionality needed to do this correctly is already there and unit tested) and will have better scaling (which should reduce algorithmic complexity from `O(n*m)` to `O(n*log(m))`.

I will post a new PR against this PR to fix the issues (with full unit tests).

e10harvey:
> I will post a new PR against this PR to fix the issues (with full unit tests).

Thanks, @bartlettroscoe !

bartlettroscoe:
@e10harvey, please review, approve, and mere https://github.com/e10harvey/TriBITS/pull/1 that will update this PR and then I can approve and merge.

bartlettroscoe (commented):
Forgot to approve this after merging in my fixes from https://github.com/e10harvey/TriBITS/pull/1

PR #353, opened 2021-02-20
Fix link in developer guide

mayrmt:
A developer-guide-internal link to another section has been fixed.


bartlettroscoe (approved):
@mayrmt, thanks for the fix!  I am surprised I am not seeing this error when generating the output PDF and HTML files.

mayrmt:
@bartlettroscoe I saw it on the html developer guide on tribits.org. I just check. It's ok now.

Issue #351, opened 2021-02-05
cdash_analyze_and_report.py: 


bartlettroscoe:
## Description

The current `cdash_analyze_and_report.py` tool is effective at notifying the user when any expected builds listed in `--expected-builds-file=<expected-builds>.csv` are missing in the table "Builds missing: bm=???" given the builds returned in the `cdash/index.php` query specified by the `--cdash-builds-filters=<build-filters>`.  However, it does not inform the user if there are builds that are returned by that query that are not expected.   So when new builds get added, it would be nice to be reminded of that and make it easy to add these new builds to the list of expected builds in the file `<expected-builds>.csv`.

## Proposed solution

To address this, would add two new arguments:

* `--show-unexpected-builds=[on|off]`:  When `on` would produce a table "Builds unexpected: bue=???" and add the number `bue=???` to the summary line.
* `--write-unexpected-builds-to-file=<unexpected-builds>.csv`: When set to non-empty, that file `<unexpected-builds>.csv` will get written that has the same fields (in the same order) as for the `--expected-builds-file=<expected-builds>.csv` file.  (If there are no expected builds, just the header line will be written with no rows below that.)

With that, a driver script would set:

```
    --show-unexpected-builds=on \
    --write-unexpected-builds-to-file=<unexpected-builds>.csv \
```

bartlettroscoe:
This would get used for the script:

* https://gitlab-ex.sandia.gov/atdm/TrilinosATDMStatus/-/blob/master/trilinos_atdm_builds_status.sh

that should have no unexpected builds.

bartlettroscoe:
With #355 merged, this Issue in "In review".

PR #355, opened 2021-02-23
List unexpected builds (#351)

bartlettroscoe:
Implements #351

### DIFF
--- a/tribits/ci_support/CDashQueryAnalyzeReport.py
+++ b/tribits/ci_support/CDashQueryAnalyzeReport.py
@@ -2592,6 +2592,104 @@ class IssueTrackerFieldError(Exception):
   pass
 
 
+# Class to report a single build-set.
+#
+# NOTE: The reason this is a class is that the cdashReportData and
+# addTestHistoryStrategy objects are set once and are used for multiple calls
+# to reportSingleBuildset().
+#
+class SingleBuildsetReporter(object):
+
+  # Constructor
+  #
+  def __init__(self, cdashReportData,
+      htmlStyle=None, htmlTableStyle=None,
+      verbose=True,
+    ):
+    self.cdashReportData = cdashReportData
+    self.htmlStyle = htmlStyle
+    self.htmlTableStyle = htmlTableStyle
+    self.verbose = verbose
+    self.groupSiteBuildNameSortOrder = ['group', 'site', 'buildname']
+
+  # Report on a given build-set and write info to self.cdashReportData
+  #
+  # Input arguments:
+  #
+  #   buildsetDescr [in] Description for the set of builds
+  #
+  #   buildsetAcro [in] Short acronym for the build set
+  #
+  #   buildsetLOD [in] List of builds in this build set
+  #
+  #   buildsetGlobalPass [in] If set to False and len(buildsetLOD) > 0, then
+  #   global pass is set to False (see below).
+  #
+  #   buildsetColor [in] Color used for the text (see cdashColorXXX() values).
+  #
+  #   buildsetColDataList [in] List of TableColumnData entries for each of the
+  #   columns to include in the table.  Default is 'None' which gives a
+  #   default set of entries.
+  #
+  #   verbose [in] If set to True then some more verbose info is printed to
+  #   STDOUT.  If False, then nothing is printed to STDOUT (which is useful
+  #   for unit testsing). The default is True.
+  #
+  # On output, self.cdashReportData data will be updated with the summary and
+  # table of this given build-set.  In particular, the following
+  # cdashReportData fields will be written to:
+  #
+  #   cdashReportData.summaryLineDataNumbersList: List will be appended with
+  #   lines for each build-set in order called.
+  #
+  #   cdashReportData.htmlEmailBodyTop: The name of the table 'buildsetDescr',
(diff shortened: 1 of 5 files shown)


bartlettroscoe:
@jmgate, @e10harvey, I will go ahead and merge and you can review post merge if you want.  That way, I can update the driver script with this.

tribits/ci_support/cdash_analyze_and_report.py
```diff
@@ -222,12 +222,26 @@ def injectCmndLineOptionsInParser(clp, gitoliteRootDefault=""):
       " file names.",
     clp )
 
+  addOptionParserChoiceOption(
```
jmgate, line 225: Probably outside the scope of this review, but this function name seems to have some redundancy in it with `Option` showing up twice.

tribits/ci_support/cdash_analyze_and_report.py
```diff
@@ -222,12 +222,26 @@ def injectCmndLineOptionsInParser(clp, gitoliteRootDefault=""):
...
+    "--list-unexpected-builds",
+    "listUnexpectedBuildsStr",
+    ("on", "off"), 1,
+    "List unexpected builds downloaded from CDash but but listed with expected builds'.",
```
jmgate, line 229: `but but`

tribits/ci_support/cdash_analyze_and_report.py
```diff
@@ -288,6 +302,11 @@ def setExtraCmndLineOptionsAfterParse(inOptions_inout):
   else:
     setattr(inOptions_inout, 'printDetails', False)
 
+  if inOptions_inout.listUnexpectedBuildsStr == "on":
+    setattr(inOptions_inout, 'listUnexpectedBuilds', True)
+  else:
+    setattr(inOptions_inout, 'listUnexpectedBuilds', False)
```
jmgate, line 308: If this were me, I'd write it as
```python
setattr(inOptions_inout, 'listUnexpectedBuilds', inOptions_inout.listUnexpectedBuildsStr == "on")
```

jmgate (commented):
Generally looks fine.  A few comments you might want to address in a future PR.  No showstoppers, though.

tribits/ci_support/CDashQueryAnalyzeReport.py
```diff
@@ -561,9 +561,38 @@ def stripWhiltespaceFromStrList(strListInOut):
...
 
 
+g_expectedBuildsCsvFileHeadersRequired = \
+  ('group', 'site', 'buildname')
+
+
 def getExpectedBuildsListfromCsvFile(expectedBuildsFileName):
   return readCsvFileIntoListOfDicts(expectedBuildsFileName,
-    ['group', 'site', 'buildname'])
+    g_expectedBuildsCsvFileHeadersRequired)
```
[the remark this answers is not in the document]
bartlettroscoe: The prefix `g_` if for "Global".  There are very few global module variables so this calls them out.

tribits/ci_support/CDashQueryAnalyzeReport.py
```diff
@@ -561,9 +561,38 @@ def stripWhiltespaceFromStrList(strListInOut):
...
+
+
+# Write list of builds from a builds LOD to a CSV file structure meant to
+# match the expected builds CSV file.
+#
+def expectedBuildsLODToCsvFileStructure(buildsLOD):
```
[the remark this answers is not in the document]
bartlettroscoe: "LOD" stands for "List Of Dicts" is a very common acronym used throughout this code.  Since Python is dynamically typed, it is sometimes nice to state the structure of the in a static way.  But generally LOD is spelled out in the function names as `ListOfDicts` so this is not consistent.  Having 'LOD' in the variable names makes more sense.  I will work to fix that.

tribits/ci_support/cdash_analyze_and_report.py
bartlettroscoe: Well, you are adding a "choice option" to an "OptionParser" object.  Since the default function is `add_option()`, this seemed like the best name.

tribits/ci_support/cdash_analyze_and_report.py
bartlettroscoe: That is a good suggestion.  I will try that out.

bartlettroscoe: That seems to work like a charm.  I have made this change in my local branch.

tribits/ci_support/cdash_analyze_and_report.py
bartlettroscoe: NOTE: With static typing in C++, you would just call this non-member function `addChoiceOption()` since the first argument would be of type `OptionParser` and that would disambiguate this function from other functions of the same name that are meant to operate on other objects.  But since Python objects are not statically typed, you need to add the class name in the non-member function name to avoid possible conflicts and avoid confusion.  (The pros and cons of dynamically typed vs. statically typed languages.)

tribits/ci_support/cdash_analyze_and_report.py
bartlettroscoe: Fixed in my local branch.

Issue #350, opened 2021-02-05
cdash_analyze_and_report.py: Allow --expected-builds-file to take list


bartlettroscoe:
CC: @e10harvey 

## Description

With the Trilinos Framework team moving to a model where they are maintaining a single `*.csv` file for the list of tests with issue trackers but multiple `*.csv` files for subsets of expected builds (and the new option `--filter-out-builds-and-tests-not-matching-expected-builds=on` implemented in PR #347) it would be very desirable to allow the option `--expected-builds-file=<file>` to accept a list of files like:

```
   --expected-builds-file=<file1>,<file2>,... \
```

Then driver scripts could run on a subset of the builds with:

```
   --expected-builds-file=<file1> \
   --tests-with-issue-trackers-file=allTestsWithIssueTrackers.csv \
   --filter-out-builds-and-tests-not-matching-expected-builds=on \
```

and

Then driver scripts could run on a subset of the builds with:

```
   --expected-builds-file=<file2> \
   --tests-with-issue-trackers-file=allTestsWithIssueTrackers.csv \
   --filter-out-builds-and-tests-not-matching-expected-builds=on \
```

but could also validate that all of the tests listed in `allTestsWithIssueTrackers.csv` match expected builds with:

```
   --expected-builds-file=<file1>,<file2> \
   --tests-with-issue-trackers-file=allTestsWithIssueTrackers.csv \
```

That way, one could:

1. Eliminate duplication between `*.csv` files listing expected builds
2.  Validate that all of the tests listed in the one `allTestsWithIssueTrackers.csv` match at least one expected build

PR #358, opened 2021-02-23
Update --expected-builds-file to accept a list (#350)

bartlettroscoe:
This is needed to eliminate duplication in the listing of expected builds in the [TrilinosATDMStatus](https://gitlab-ex.sandia.gov/atdm/TrilinosATDMStatus) driver scripts.  Also, this will make it clear if any new unexpected builds show up on CDash for the full driver script.

### DIFF
--- a/tribits/ci_support/CDashQueryAnalyzeReport.py
+++ b/tribits/ci_support/CDashQueryAnalyzeReport.py
@@ -565,9 +565,19 @@ g_expectedBuildsCsvFileHeadersRequired = \
   ('group', 'site', 'buildname')
 
 
-def getExpectedBuildsListfromCsvFile(expectedBuildsFileName):
+def getExpectedBuildsListOfDictsfromCsvFile(expectedBuildsFileName):
   return readCsvFileIntoListOfDicts(expectedBuildsFileName,
     g_expectedBuildsCsvFileHeadersRequired)
+
+
+def getExpectedBuildsListOfDictsFromCsvFileArg(expectedBuildsFileArg):
+  expectedBuildsLOD = []
+  if expectedBuildsFileArg:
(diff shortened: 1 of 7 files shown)


tribits/ci_support/CDashQueryAnalyzeReport.py
```diff
@@ -565,11 +565,21 @@ def stripWhiltespaceFromStrList(strListInOut):
...
 
-def getExpectedBuildsListfromCsvFile(expectedBuildsFileName):
+def getExpectedBuildsListOfDictsfromCsvFile(expectedBuildsFileName):
   return readCsvFileIntoListOfDicts(expectedBuildsFileName,
     g_expectedBuildsCsvFileHeadersRequired)
 
 
+def getExpectedBuildsListOfDictsFromCsvFileArg(expectedBuildsFileArg):
+  expectedBuildsLOD = []
+  if expectedBuildsFileArg:
```
jmgate, line 575: Apparently the preference in the Python community would be to use
```python
if expectedBuildsFileArg is not None:
```

This was news to me when I found out about it a while ago, but having adopted the practice, I do appreciate how it eases reading the code.

bartlettroscoe: Not sure I agree with this totally with this.  The statement `if <var>` evaluates to false for the empty string is "" instead of `None` and other conditions as well (like a list with zero elements).  In fact, in this case I suspect that `expectedBuildsFileArg` is actually the empty string instead of `None` but the logic  `if <var>`  applies in either case.

So in this case, I suspect the check would be `if expectedBuildsFileArg is not ""` or the equivalent.

jmgate: Ah, right, and in that case you'd want to go with `!=` in place of `is not`, I think.

PR #362, opened 2021-03-17
tribits/ci_support: Remove soft hyphens

e10harvey:
- Remove soft hyphens from htmlBody in createHtmlMimeEmail.

During triaging we often want to copy long build names or test names from the email bodies into CDash queries but run into issues. It turns out that our browsers don't render these soft hyphens but these hidden formatting characters cause the CDash queries to fail.

CC: @ZUUL42, @prwolfe, @jwillenbring, @william76

### DIFF
--- a/tribits/ci_support/CDashQueryAnalyzeReport.py
+++ b/tribits/ci_support/CDashQueryAnalyzeReport.py
@@ -3022,6 +3022,10 @@ def createHtmlMimeEmail(fromAddress, toAddress, subject, textBody, htmlBody):
   # through the gateway when html text lines are longer than 1000 characters.
   msg['Content-Transfer-Encoding'] = "quoted-printable"
 
+  # Remove hidden soft hyphens from htmlBody so triagers can easily copy paste
+  # long build and test names into CDash queries.
+  htmlBody = htmlBody.replace('&shy;','')
+
   # Record the MIME types of both parts - text/plain and text/html.
   part1 = MIMEText(textBody.encode('utf-8'), 'plain', 'utf-8')
   part2 = MIMEText(htmlBody.encode('utf-8'), 'html', 'utf-8')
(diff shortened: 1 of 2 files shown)


bartlettroscoe (changes requested):
@e10harvey, if you remove those soft hyphens, then that damages the formatting of the tables when there a long names where it can break on underscores `_`.  I added these soft hyphens after a lot of experimentation and experience to allow breaks along `_` chars.  It is better to get these names by other means (like from the CDash query URLs).

But if you want to remove the soft hyphens, then we need to add a command-line option to do so and then the user can decide if they want to have better formatted and easier to read tables or not have soft hyphens.

Can you add a command-line option to remove the soft hyphens with the default off?

e10harvey:
Thanks for your feedback, @bartlettroscoe. I assumed the soft hyphens got added unintentionally. For what it's worth, the table format looks fine both with and without the soft hyphens on my end.

bartlettroscoe:
@e10harvey, this is close.  I will post a PR against this PR with some final fixups to make more consistent with the rest of the code.

But I will note that there is no automated testing for this new option an behavior.  Therefore, if future changes to this code brakes this, then the tests will not show that.  If you are comfortable with that then so am I.

bartlettroscoe:
@e10harvey, the staggered PR https://github.com/e10harvey/TriBITS/pull/2 contains the changes that address the issues.  If you merge that PR to your branch (which will update this PR branch), then I will approve and merge this PR to 'master'.

bartlettroscoe:
@e10harvey, FYI, I did not manually test running with the new option enabled when I posted https://github.com/e10harvey/TriBITS/pull/2.  Can you verify that it still behaves the way you expect?  If you want to add an automated test for this then that will take some refactoring.

Otherwise, let me know if you me to approve and merge this branch as-is.

e10harvey:
> Can you verify that it still behaves the way you expect?

Yes. The manual tests passed.

An automated test is not a high priority right now.

bartlettroscoe (approved):
I will merge and push with the checkin-test.py script locally with full testing.  (Would be nice to set up GitHub Actions tests at some point here.)

Issue #136, opened 2016-07-15
TRIBITS_ADD_ADVANCED_TEST() functionality, performance, and error checking upgrade


bartlettroscoe:
There are a few related things that need to be addressed with `TRIBITS_ADD_ADVANCED_TEST()`:
* There is currently a hard-coded restriction of 20 test cases (`TEST_0` ... `TEST_19`) (and people have tried to create more test cases).
* There is currently a potential performance problem with a lot of keyword arguments and long argument lists (O(numKeywords * numTotalArguments) (see email chain below).  This need to be characterized through performance tests.
* Non-consecutive test cases are not caught and remaining test cases are ignored (see #3130).
* Overall arguments interspersed with `TEST_<IDX>` blocks result in strange errors (see comment:2)
* The numbers of argument values in keyword arguments is not being checked (e.g. that "TIMEOUT" must have exactly one value)

I think that a rewriting of the current argument parsing code that uses PARSE_ARGUMENTS() can address all of the above issues in one shot.  I will outline the proposed algorithm (and compare it to the existing algorithm) in a following comment.   However, the changed algorithm will have performance implementations so I would like to set up a simple performance test to see what impact my changes will have.

Some other issues that would be nice to address as well:
* The number of MPI processes is always set to the default, even when no MPI executables are added.  This tells CTest that there are 4 processes needed when only 1 process is needed.

**Email chain:**
----
From: Bartlett, Roscoe A. 
Sent: Monday, March 10, 2014 10:36 AM
To: 'Robert Salko'
Subject: RE: tribits AddAdvancedTest

Bob,

I will see about removing the restriction on the number of test cases today.  However, there are some performance implications that I need to look into so it may take some time to set up good performance testing cases and see what the real impact is.  After looking at the implementation of TRIBITS_ADD_ADVANCED_TEST() and the helper script it uses PARSE_ARGUMENTS(), I fear that very long argument lists with lots of list arguments might be expensive (it looks like the complexity is O(numTestCases * numTotalArgs), which will look like quadratic complexity) so I fear this might get expensive in a case like yours.  This is all implemented in CMake (for portability) and there are limits that what you seem to be able to do with the language.

However, if this is a common set of tasks such that a user might want to do this, you might consider writing some reusable, tested, scripting code (i.e. in Python) to do some of these sub-tasks.  You will want to unit test these little scripts so you can trust them.  You can then just plug these into the TRIBITS_ADD_ADVANCED_TEST() command.

[...]

PR #365, opened 2021-04-15
TRIBITS_ADD_ADVANCED_TEST(): Allow increase in max num TEST_ <idx> blocks (#136)

bartlettroscoe:
This will fix the problem described in https://github.com/TriBITSPub/TriBITS/issues/136#issuecomment-818921516 from @glhenni.  This is not the full fix for #136 but it will resolve that use case.


bartlettroscoe:
I will go ahead and test and merge this with the checkin-test.py script and a post-merge review can be done.

PR #366, opened 2021-04-19
CI: draft action yaml

fnrizzi:
Commits squashed into one, please note that the CI still fails as anticipated


bartlettroscoe:
@marcinwrobel1986, basically, we need to update the GitHub action to run the command:

```
$ ctest -V -S ../TriBITS/cmake/ctest/travisci/ctest_trusty_serial_debug.cmake
```

instead of separate 'cmake', 'make', and 'ctest' commands same as for:

* https://github.com/TriBITSPub/TriBITS/blob/master/.travis.yml

bartlettroscoe:
This was duplicated by #389 then #401.  Thanks for your initial work on this @fnrizzi!

Issue #377, opened 2021-06-11
Provide feedback on TriBITS package dependencies data-structures


bartlettroscoe:
@fnrizzi, @MikolajZuzek, @marcinwrobel1986, @keitat 

**Parent Issue:** #63

## Description

This story will be to get feedback from NextGen Analytics staff on the basic data-structures and functions for the TriBITS packages dependencies system that is so critical to managing the complex internal and external dependencies of a large CMake project.  It is critical to understand these basic data-structures and the code that creates and manipulates them in order to have anything that resembles TriBITS.  

**NOTE**: This code has nothing to do with the internally called commands like `tribits_project()`, `tribits_package()`, `tribits_add_library()` or `tribits_add_executable()` .   That is, this code is completely 100% orthogonal to how targets are created in the CMake project.  It only determines the set of external and internal packages that will be enabled and the order and location of the CMakeLists.txt file that need to be included with `add_subdirectory()` commands.

The package dependency data-structures themselves (which are currently in a state of being refactored as per #63) are described in:

* https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-system-data-structures-and-functions

(NOTE: Some of the variables like `${PACKAGE_NAME}_PACKAGE_STATUS` don't exist yet.  This documentation is part of the planning for the refactoring to support #63.)

The code that reads in the `*.cmake` fragment files and creates the internal data-structures for the list of packages and their dependencies is rooted in the macro [`TRIBITS_READ_ALL_PROJECT_DEPS_FILES_CREATE_DEPS_GRAPH()`](https://tribits.org/doc/TribitsDevelopersGuide.html#tribits-read-all-project-deps-files-create-deps-graph)  and its function call graph is given in:

* https://tribits.org/doc/TribitsDevelopersGuide.html#function-call-tree-for-constructing-package-dependency-graph

The `<filename>:<line>` location of each of these macros and functions are given in the extracted documentation blocks.  The bulk of the code in this call graph are defined in just a few `*.cmake` files (as you can see in the documentation).

A CMake -P utility script that runs this code to create and export the dependency graph for any TriBITS project as an XML file is given in:

* https://github.com/TriBITSPub/TriBITS/blob/master/tribits/ci_support/TribitsDumpDepsXmlScript.cmake

[...]

PR #388, opened 2021-07-15
#377: Fixing typos

marcinwrobel1986:
Addresses #377

### This PR:
- fixes text typos in:
  - tribits/core/package_arch/TribitsReadAllProjectDepsFilesCreateDepsGraph.cmake
  - tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
  - tribits/core/package_arch/TribitsSystemDataStructuresMacrosFunctions.rst
- there are still typos in functions and variables names, but I would leave that to @bartlettroscoe

### DIFF
--- a/tribits/core/package_arch/TribitsReadAllProjectDepsFilesCreateDepsGraph.cmake
+++ b/tribits/core/package_arch/TribitsReadAllProjectDepsFilesCreateDepsGraph.cmake
@@ -74,7 +74,7 @@ macro(tribits_read_all_project_deps_files_create_deps_graph)
 
   tribits_config_code_start_timer(SET_UP_DEPENDENCIES_TIME_START_SECONDS)
 
-  tribits_read_defined_external_and_intenral_toplevel_packages_lists()
+  tribits_read_defined_external_and_internal_toplevel_packages_lists()
 
   tribits_read_deps_files_create_deps_graph()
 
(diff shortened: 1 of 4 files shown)


bartlettroscoe:
> there are still typos in functions and variables names, but I would leave that to @bartlettroscoe

@marcinwrobel1986, please go ahead and make any name changes you want in this PR.  I will make sure nothing breaks before the merge.

marcinwrobel1986:
@bartlettroscoe Hello Ross, I think it's ready to be reviewed

bartlettroscoe:
> @bartlettroscoe Hello Ross, I think it's ready to be reviewed

Thanks @marcinwrobel1986!  I will take care of this after #389 (which will test this automatically :-))

bartlettroscoe (approved):
I suspect this may conflict with PR #403 but I will take care of those conflicts after this merges.

I really need to run the spell checker on this code base more often to catch things like this.  (But I am not sure it would catch misspellings inside of longer macro/function names like this.)  I guess I just need to be more careful when first writing these to separate out the words and spell check them individually and then concatenate them into the macro/function name.

marcinwrobel1986:
> I suspect this may conflict with PR #403 but I will take care of those conflicts after this merges.

Thanks Ross!

> I really need to run the spell checker on this code base more often to catch things like this. (But I am not sure it would catch misspellings inside of longer macro/function names like this.) I guess I just need to be more careful when first writing these to separate out the words and spell check them individually and then concatenate them into the macro/function name.

An IDE with spell check wouldn't be a bad choice :)

bartlettroscoe:
> An IDE with spell check wouldn't be a bad choice :)

Yea, it may be time to look into that, even if it is just inside of emacs.  (Getting an IDE to work on all of the various SNL machines is very difficult.)

marcinwrobel1986:
Hello Ross, I don't use emacs, but I have found something. Maybe it's worth to give it a try:
[Spell Checking in Emacs](https://www.tenderisthebyte.com/blog/2019/06/09/spell-checking-emacs/)

Issue #386, opened 2021-06-27
Build TriBITS Documentation with Sphinx and deploy to the web


bartlettroscoe:
CC: @marcinwrobel1986, @fnrizzi 

## Description

As discussed at the prior meeting on 6/24/2021, this Issue is to investigate the building and rendering of TriBITS documentation written in reStructuredText using Sphinx and deploy somehow to the web.

There are three documents that are generated using as reStructuredText files using the script:

```
$ cd TriBITS/
$ cd tribits/doc/
$ ./build_docs.sh
```

All the details of what gets built and how can be found in that script and the scripts and tools it calls.

To build the documentation locally, one needs to have docutils installed.  But even without docutils installed, these scripts will still build everything needed to for the assembled files:

```
tribits/doc/guides/maintainers_guide/TribitsMaintainersGuide.rst
tribits/doc/guides/users_guide/TribitsUsersGuide.rst
tribits/doc/build_ref/TribitsBuildReference.rst
```

The script:

```
tribits/doc/publish_docs.sh
```

is used to publish the generated *html files on the tribits.org site (but you can run it locally to test it out).

## Tasks

* [x] Implement sphinx build of the documents ... See PR #387 
* [x] Deploy sphinx documents as a GitHub pages site ... See the site https://tribitspub.github.io/TriBITS/ served off of of the branch https://github.com/TriBITSPub/TriBITS/tree/deploy-doc-site
* [x] Update tribits.org to point  to new site and new in-progress Sphinx-generated documents
* [ ] Create follow-on GitHub Issue listing remaining issues that need to be resolved before abandoning the DocUtils-generated documentation and fully adopting the Sphinx-generated documentation.

bartlettroscoe:
CC: @fnrizzi, @keitat 

@marcinwrobel1986, as we discussed at last Thursday's meeting, this is the issue to investigate using Sphinx and readthedocs.com to improve the rendering of the TriBITS documentation currently hosted at tribits.org.

marcinwrobel1986:
please find a new documentation under this address:
https://marcinwrobel1986.github.io/TriBITS/

What was done:
- a python script for combining together documentation for Sphinx
- Sphinx configuration with the landing page
- CI (github actions) for generating and deploying docs to GitHub pages

Potential threats:
- script `build_docs.sh` needs to be run from it's directory
- it's hard to destinguish any sensible debug method at the current docs generating process
[...]

PR #391, opened 2021-07-16
Fix up build_docs.sh for Sphinx doc generation (#386)

bartlettroscoe:
* Makes so `build_docs.sh` can run from any base directory
* Add argument `--skip-final-generation`


bartlettroscoe:
@marcinwrobel1986, this will allow you to add `--skip-final-generation` the call of `tribits/doc/build_docs.sh`.

PR #402, opened 2021-08-02
Revert some incorrect uppercase->lowercase changes

gsjaardema:
I think that the two -L -> -l and -I -> -i changes are incorrect and they should be -L and -I


bartlettroscoe (approved):
Yup, these would have matched the replacement patterns for a cmake command call.  That is unfortunate.  Come to think of it, I don't think valid CMake code would ever have a char like `-` right before a command-call.

As the conversion scripts should be run on Trilinos and other TriBIT-based projects at some point, we should likely try to catch cases like this or provide a mode where more conservative replacements are done.

PR #403, opened 2021-08-02
Spelling fixes

gsjaardema:
Fix several incorrectly spelled words...

Command used was `codespell -i 3 -w -L inout,ned,aci` which resulted in few false-positives.


gsjaardema:
Fixes #404

tribits/core/package_arch/TribitsAddTest.cmake
```diff
@@ -336,7 +336,7 @@ include(TribitsAddTestHelpers)
 #     send to stdout.  Otherwise, the test will fail.  This is set using the
 #     built-in CTest property ``PASS_REGULAR_EXPRESSION``.  Consult standard
 #     CMake documentation for full behavior.  TIPS: Replace ';' with '[;]' or
-#     CMake will interpret this as a array element boundary.  To match '.',
+#     CMake will interpret this as an array element boundary.  To match '.',
```
bartlettroscoe, line 339: Wow, codespell can take into the account the context of the sentence as well to suggest updated works.  It is not just looking at works in isolation.  Very nice!

tribits/doc/README.DEPENDENCIES
```diff
@@ -0,0 +1,251 @@
+################################################################################
```
bartlettroscoe, line 1: I think this file may be removed in the updated version TriBITS 'master'?

tribits/doc/developers_guide/generate-dev-guide.sh
```diff
@@ -0,0 +1,193 @@
+#!/bin/bash
```
bartlettroscoe, line 1: This file got moved and renamed in the updated version of TriBITS.

bartlettroscoe (commented):
@gsjaardema, thanks for posting this and taking the time to do this.  I did not know about the tool [codespell](https://github.com/codespell-project/codespell) but it seems pretty useful.  I will see about installing it and see if I can integrate it into my development processes.

The only issue I see is that it looks like this was run on an older version of TriBITS 'master' and some of the files have changed names and have been moved around.  Therefore, I may need to push a few more commits to this branch to address these things to make sure.

PR #394, opened 2021-07-20
DO NOT MERGE: Show TriBITS test failures with CMake 3.21.0 that don't occur with CMake 3.17.5 (#363)

bartlettroscoe:
This PR is used to demonstrate the different behavior of CMake 3.21.0 vs. CMake 3.17.5 with the TriBITS test suite.  It uses a GitHub Actions driver that posts to CDash two otherwise identical builds that differ only in the CMake version that they use.  These test failures are blocking the deployment of the GitHub Actions testing in PR #393 to implement #363.

Issue #363, opened 2021-03-23
Set up TriBITS testing using GitHub Actions


bartlettroscoe:
CC: @keita, @fnrizzi, @marcinwrobel1986, @MikolajZuzek 

With a lot of work on TriBITS going on and coming up (i.e. #63, #299, etc.) we need solid testing for TriBITS.  I have solid testing of TriBITS on local machines set up using the checkin-test.py script (see checkin-test-crf450.sh and those it calls) but we really need to have this for PR testing as well.  In the past, I did that with Travis CI.  But for reasons that I can't go into here, Travis CI is not a viable option going forward.

The obvious choice for this would seem to be to use GitHub Actions (GHA) for this.  But I don't know anything about GHA so this will be a learning curve.

For dependencies I need:

* CMake 3.17
* C++, C, and Fortran compilers
* Python 2.7

Other dependencies that would be nice to have:

* CMake current release (current CMake 3.19 but soon to be 3.20)
* Python 3.x
* Ninja 1.10+

And this would post to CDash at:

* https://testing.sandia.gov/cdash/viewSubProjects.php?project=TriBITS

At a minimum, there needs to be a build with:

* CMake 3.17 + Python 2.7 (or 3.x)

Other builds that would be nice to have include:

* CMake current + Python 3.x (or 2.7 if base build is using 3.x)
* CMake 3.17 + Ninja 1.10
* CMake 3.17 + Ninja 1.10 (no Fortran)

## Tasks:

* [x] Scope out new GitHub Actions driver (see PRs #366 and #389)
* [x] Create new branch '363-github-actions-tests' based off '377-provide-feedback-on-TriBITS-package-dependencies-data-structures-CI', rebase, clean up commits and create new PR .
* [x] Add install of docutils to github actions job and enable build_docs test.
* [x] Create file `tribits_cmake-3.21.0_makefiles_python-3.8_g++-9_tweaks.cmake` disables for cmake-3.21.0 GHA build to disable tests that fail with CMake 3.20.
* [x] Add new triggers `workflow_dispatch` and `schedule`.
* [x] Determine what branch the GitHub Action jobs was invoked on and send jobs to CDash to different CDash groups based on the event trigger: PRs and manual triggers to 'Experimental', pushes to 'master' to 'Continuous', cron jobs to 'Nightly' (see GITHUB_EVENT_NAME).
[...]

PR #406, opened 2021-08-12
Fix tribits_ctest_driver() package-by-package mode for CMake 3.19+ (#363, #394)

bartlettroscoe:
Fixes one of the tests shown and discussed in #394 (the rest will be fixed by CMake 3.21.2).  This is related to #363.

See the commits for more details.

NOTE: I listed this PR as a bug even though it was triggered by a change in CMake.  This use case should have never relied on cmake_build() building a missing target and not returning an error.  So this really was a bug in TriBITS logic.


bartlettroscoe:
I tested this locally as well and it looks good.  This is ready to merge.

@marcinwrobel1986, you can review post-merge if you want.  This is about 95% comments and tests and very little production coded (as is often the case in tricky situations like this).

PR #412, opened 2021-09-02
DO NOT MERGE: Some experiments with targets for handling external libraries (#299)

bartlettroscoe:
There are a serious of failed experiments for using IMPORTED and INTERFACE libraries for dealing with externally built libraries and turning them into CMake targets.  The purpose is to experiment with a few strategies that might be used for TPLs/external packages as part of #299 and #63.  The results are not good.  I am going to need some help from Kitware on this one.  But at least this PR branch and the commits document the experiments and what some of the issues are.  (You have to build the TribitsExampleProject locally manually to reproduce the issues but that is easy to do.)


bartlettroscoe:
FYI: For the last testing iteration [here](https://github.com/TriBITSPub/TriBITS/actions/runs/1209839375) for the commit e1fc467a5a7f50f30011e3d63956050cfb663c40, a bunch of those tests fail because I allowed the WrapExternal package to be enabled when package-config files are enabled (which is kind of a hack but works).

bartlettroscoe:
I don't need to keep this open anymore.  I have implemented all of this as part of #299 (most recently with PR #436).

PR #413, opened 2021-09-02
Change internal TriBITS target_link_libraries() to PUBLIC (#299)

bartlettroscoe:
This should fix the propagation of Kokkos compiler flags added on Kokkos library targets using the INTERFACE_COMPILE_OPTIONS target property.   See the commit msg for more details.

Part of #299.  Much more to come in other topic branches and PRs.

### DIFF
--- a/tribits/core/package_arch/TribitsAddExecutable.cmake
+++ b/tribits/core/package_arch/TribitsAddExecutable.cmake
@@ -175,7 +175,7 @@ include(CMakeParseArguments)
 #   ``TESTONLYLIBS <lib0> <lib1> ...``
 #
 #     Specifies extra test-only libraries defined in this CMake project that
-#     will be linked to the executable using ``target_link_library()``.  Note
+#     will be linked to the executable using ``target_link_libraries()``.  Note
 #     that regular libraries (i.e. not ``TESTONLY``) defined in the current SE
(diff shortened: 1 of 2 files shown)


bartlettroscoe:
@marcinwrobel1986, there is not really much to review here.  The real work will be testing with Trilinos.  I will create the PR to update TriBITS in Trilinos next.

marcinwrobel1986:
@bartlettroscoe I have just built Trilinos, but then I figure out, the TriBITS changes are not there yet :)

bartlettroscoe:
> @bartlettroscoe I have just built Trilinos, but then I figure out, the TriBITS changes are not there yet :)

@marcinwrobel1986, I will post the Trilinos PR shortly.  But note that you can clone the TriBITS repo under the Trilinos repo and then use that TriBITS instead of the one in Trilinos when you configure Trilinos.  See:

* https://docs.trilinos.org/files/TrilinosBuildReference.html#select-different-tribits-implementation

(Looks like that documentation item needs to be updated.  I will do that in my Trilinos PR as well.)

bartlettroscoe:
@marcinwrobel1986, hold up, I am getting configure and (after fixing those) build errors with Trilinos after using this version of TriBITS.  Let me worth through the issues.

PR #414, opened 2021-09-09
Add TribitsExampleApp and tests to test TriBITS-generated <Package>Config.cmake files (#299)

bartlettroscoe:
Working towards #299

This PR adds the new example/test raw CMake project TribitsExampleApp. It define a C++ application that calls some functions defined in TribitsExampleProject (TribitsExProj) and pulls in what it needs through `<Package>Config.cmake` files.  This APP can conditionally use different packages/components from TribitsExampleProject and can pull in either `find_package(TribitExProj REQUIRED COMPONENTS ... )` at the project level or pull in individual packages with individual `find_package(<Package> REQUIRED)` calls.

The primary purpose of CMake project is to test the `<Package>Config.cmake` files currently being generated by TriBITS.  I added several TriBITS tests with this for these files both being installed and those being placed in the build tree.

With these tests in place, I feel fairly confidence to refactor the existing library and exec targets for internal packages to use modern modern CMake for handling include dirs and other bits of info (see #299).

### DIFF
--- a/test/core/ExamplesUnitTests/CMakeLists.txt
+++ b/test/core/ExamplesUnitTests/CMakeLists.txt
@@ -4540,6 +4540,93 @@ TribitsExampleApp_ALL_ST_test(ByPackage STATIC)
 TribitsExampleApp_ALL_ST_test(ByPackage SHARED)
 
 
+function(TribitsExampleApp_ALL_ST_buildtree_test sharedOrStatic)
+
+  if (sharedOrStatic STREQUAL "SHARED")
+    set(buildSharedLibsArg -DBUILD_SHARED_LIBS=ON)
+  elseif (sharedOrStatic STREQUAL "STATIC")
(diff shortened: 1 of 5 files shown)


marcinwrobel1986 (approved):

marcinwrobel1986:
Hello @bartlettroscoe I was able to build and test the changes. All test regarding this example app have passed.

bartlettroscoe:
This new example could likely use a TribitsExampleApp/README.md file to explain what I write [above](https://github.com/TriBITSPub/TriBITS/pull/414#issue-730521164) but I can do that later.  This example may change some as we complete #299 and #63 so we might as well wait to document it after that.  (But if we do those refactorings correctly to maintain backward compatibility, this example and the tests using it should not have to change at all.  That is the whole point of having this example and the TriBITS tests based on it actually.)

I will go ahead and merge this so that I can do the follow-on work.

Issue #415, opened 2021-09-09
Address duplicate nightly GitHub Actions jobs


bartlettroscoe:
## Next Action Status

PR #416 merged to 'master' eliminates submits to CDash from forks.  Next:  Have @gsjaardema update 'master' in his fork https://github.com/gsjaardema/TriBITS (i.e. merge https://github.com/gsjaardema/TriBITS/pull/1) ...

## Description

Somehow there are duplicate nightly builds being submitted by GitHub Actions for the `.github/workflows/tribits_testing.yml` file.  One set of builds comes from that file on the 'master' branch and the other seems to be coming from an older version of that file before the commit a45727a9434e4de52b42ecf751c037de1bdecaee that upgraded from CMake 3.21.0 to 3.21.2.  You can see evidence of this each day at ???, for example, today 2020-09-09 showing:

![image](https://user-images.githubusercontent.com/1959736/132711164-824132a2-b9bd-4f5a-9ee4-5bb8ba822a30.png)

Notice the separate builds `tribits_cmake-3.21.0_makefiles_python-3.8_g++-9` and `tribits_cmake-3.21.2_makefiles_python-3.8_g++-9`.  The build `tribits_cmake-3.21.0_makefiles_python-3.8_g++-9` does not show an "Update" column so that shows it is from a version of the file that is pretty old.

Now, the GitHub Actions in this repo shown [here](https://github.com/TriBITSPub/TriBITS/actions/workflows/tribits_testing.yml?query=event%3Aschedule) only shows the jobs running once a day showing:

![image](https://user-images.githubusercontent.com/1959736/132711868-fa61512f-f20b-40d6-a5a0-cfac339b2f4c.png)

which shows it running only once a day.

We need to remove this phantom extra nightly build because it is messing up the nightly build results on CDash.

bartlettroscoe:
One clue we can see is that the TriBITS configure captures the git SHA1 as you can see [here](https://testing.sandia.gov/cdash/build/9890817/configure) for example for the build `tribits_cmake-3.21.0_makefiles_python-3.8_g++-9` last night showing:

```
*** Base Git Repo: TriBITS
cab6dee [Mon Jul 26 14:34:24 2021 -0400] <rabartl@sandia.gov>
Remove cache and cleanup
```

while the build ` tribits_cmake-3.21.2_makefiles_python-3.8_g++-9` last nightly configure results [here](https://testing.sandia.gov/cdash/build/9890759/configure) showed:

```
[...]

PR #416, opened 2021-09-10
Only run GitHub Actions testing in main repo (#415)

bartlettroscoe:
This should eliminate problems with forks of TriBITS submitting to CDash when they update 'master' and have GitHub Actions enabled in their fork (see #415).


bartlettroscoe:
I will go head and merge this to further test this on 'master' and I will merge 'master' to 'master' in my fork to verify it does not submit to CDash.

PR #418, opened 2021-09-10
Enable SimpleTpl to TribitsExampleApp tests (#299)

bartlettroscoe:
This will allow testing the generation of `<tplName>Config.cmake` files with modern CMake targets for external packages/TPLs.  I will need more extensive tests for TPLs but this allows that refactoring to occur with some testing.

NOTE: With this PR, we now have test cases using `<Package>Config.cmake` and `<Project>Config.cmake` files that have two TPLs enables (with that is headers only, and one with a header file and a library).

NOTE: This changed what the `SimpleCxx::deps()` function returns to include calling functions from the TPLs so that required updating several tests in TriBTS.  These are all now much better tests of the usage of a TPL in TriBITS (but very limited obviously because there are just two TPLs here).

See the commit messages for more details.

### DIFF
--- a/.github/workflows/tribits_testing.yml
+++ b/.github/workflows/tribits_testing.yml
@@ -80,6 +80,7 @@ jobs:
         run: |
           cd ..
           env CTEST_BUILD_NAME_SUFFIX=${{ github.event.inputs.ctest_build_name_suffix }} \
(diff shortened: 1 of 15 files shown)


bartlettroscoe:
I will go ahead and merge this PR so I can move on to the next thing.  This can always be reviewed post-merge.

Issue #427, opened 2021-11-03
tests suddenly failing in LAPACKConfig.cmake


glhenni:
I believe something in TriBITs has broken our code. Last night all our tests failed during the configure phase with errors from tribits shown below. We're using cmake 3.19.2
```
CMake Error at /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TEST.COV/cmake_packages/LAPACK/LAPACKConfig.cmake:17 (add_library):
  add_library cannot create imported target "LAPACK::dl" because another
  target with the same name already exists.
Call Stack (most recent call first):
  /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TriBITS/tribits/core/package_arch/TribitsTplFindIncludeDirsAndLibraries.cmake:686 (include)
  /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TriBITS/tribits/common_tpls/FindTPLLAPACK.cmake:107 (tribits_tpl_find_include_dirs_and_libraries)
  /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TriBITS/tribits/core/package_arch/TribitsProcessEnabledTpl.cmake:106 (include)
  /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TriBITS/tribits/core/package_arch/TribitsGlobalMacros.cmake:1565 (tribits_process_enabled_tpl)
  /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TriBITS/tribits/core/package_arch/TribitsProjectImpl.cmake:196 (tribits_process_enabled_tpls)
  /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TriBITS/tribits/core/package_arch/TribitsProject.cmake:93 (tribits_project_impl)
  CMakeLists.txt:61 (TRIBITS_PROJECT)


CMake Error at /scratch/charonops/jenkins/workspace/ASCIC_Build_Farm_CDE-GNU-cov/TEST.COV/cmake_packages/LAPACK/LAPACKConfig.cmake:23 (add_library):
  add_library cannot create imported target "LAPACK::m" because another
  target with the same name already exists.
Call Stack (most recent call first):
[...]

PR #431, opened 2021-11-12
Address FindTPL<tplName>.cmake inner find_package() calls finding TriBITS-installed <tplName>Config.camke files and add beginning of new examples '2' (#299, #427)

bartlettroscoe:
This PR first adds a failing tests that basically reproduces the problems highlighted in bug report #427 in commit fb79c7f.  The (what should be) super robust solution for this given in commits b4a6628ba2d1b3e79b14e0cfda78fc675051916c and b4a6628ba2d1b3e79b14e0cfda78fc675051916c and is is explained at the bottom of the TriBITS Users Guide section "How to use find_package() for a TriBITS TPL":

* https://github.com/TriBITSPub/TriBITS/blob/8fe2a49f6f045aa95801dd40aab0ae4ef7688a35/tribits/doc/guides/TribitsGuidesBody.rst#how-to-use-find_package-for-a-tribits-tpl

which is:

> Therefore, to avoid having an installed TriBITS-generated ``HDF5Config.cmake``
file, for example, being found by the inner call to ``find_package(HDF5 ...)``
in the file ``FindTPLHDF5.cmake`` (which would be disastrous), TriBITS employs
two safeguards.  First, TriBITS-generated ``<tplName>Config.cmake`` files are
placed into the build directory ``<buildDir>/external_packages/`` and
installed into the directory ``<installDir>/lib/external_packages/`` so they
will not be found by default when ``<buildDir>/cmake_packages`` and/or
``<installDir>``, respectively, are added to ``CMake_PREFIX_PATH``.  Also,
even if ``<installDir>/lib/external_packages`` or
``<buildDir>/external_packages`` do get added to the search path somehow
(e.g. through ``CMAKE_INSTALL_PREFIX``), the TriBITS framework will set the
variable ``TRIBITS_FINDING_RAW_<tplName>_PACKAGE_FIRST=TRUE`` before including
``FindTPL<tplName>.cmake`` and there is special logic in the TriBITS-generated
``<tplName>ConfigVersion.cmake`` file that will set
``PACKAGE_VERSION_COMPATIBLE=OFF`` and result in ``find_package(<tplName>)``
not selecting ``<tplName>Config.cmake``.  (It turns out that CMake's
``find_package(<Package>)`` command always includes the file
``<Package>ConfigVersion.cmake``, even if no version information is passed to
the command ``find_package(<Package>)``.  This allows special logic to be
placed in the file ``<Package>ConfigVersion.cmake`` to determine if
``find_package(<Package>)`` will select a given ``<Package>Config.cmake`` file
that is in the search path based on a number of different criteria such as in
this case.)

Also in this PR:

* 6ff0323: Provide a first example of a `FindTPL<tplName>.cmake` file that writes a custom `<tplName>Config.cmake` file that uses `find_dependency()` with `<tplName>_DIR` set to find upstream external non-TriBITS package.
* 6f0e888: Added error message for `find_pacakge(<Project> REQUIRED COMPONENTS <c0> <c1>)` describing what packages were not present when finding a project's `<Project>Config.cmake` file with components and a test to ensure the right error message gets printed.
* Created the beginnings of `TribitsExampleProject2` and `TribitsExampleApp2` that will allow deeper testing with dependent TPLs so that I can finish implementing the move to modern CMake targets and stripping the old complex TriBITS logic (#299).


bartlettroscoe:
I will go ahead and do the merge and any reviews can take place post-merge.  I need to address a known problem with MS Windows and then get this PR added to the TriBITS snapshot for https://github.com/trilinos/Trilinos/pull/9894.

Issue #433, opened 2021-12-09
Address issues from TriBITS snapshot into Trilinos from PR trilinos/Trilinos#9894 (#299)


bartlettroscoe:
**Parent Issue:**
* #299

**Child Issues:**
* #443

## Description

This issue is to track TriBITS work to address the problems that were identified with the merge of the updated TriBITGS snapshot into Trilinos with the merge of Trilinos PR trilinos/Trilinos#9894 which lead the the revert PR trilinos/Trilinos#9977.  The fixes will be snapshotted to the branch in Trilinos PR trilinos/Trilinos#9978.  All of this is related to #299.

## Tasks

* [x] 1. Issues causing Albany configure failures from trilinos/Trilinos#9972 ... See PR #435
* [x] 2. Handling of general linker options in `TPL_<tplName>_LIBRARIES` like `-mkl` described in https://github.com/trilinos/Trilinos/pull/9978#issuecomment-982834778
* [x] 3. Handling of library names in `TPL_<tplName>_LIBRARIES` described in https://github.com/trilinos/Trilinos/pull/9978#issuecomment-982166363
* [x] 4. Bring back DEPRECATED non-namespace IMPORTED targets (see https://github.com/TriBITSPub/TriBITS/issues/299#issuecomment-1011558738).
* [x] 5. Reproducing and fixing issues with Nalu (see https://github.com/trilinos/Trilinos/pull/9978)
* [x] 6. Reproducing and determine how to fix issues with SPARC (see https://github.com/trilinos/Trilinos/pull/9978)
* [x] 7. Document change to `-isystem` for IMPORTED targets and workarounds (see #443)
* [x] 8. Update TriBITS snapshot in https://github.com/trilinos/Trilinos/pull/9978 and update Trilinos release notes and build reference documents (see #443)
* [ ] 9. Get Trilinos PR trilinos/Trilinos#9978 merged to Trilinos 'develop'

bartlettroscoe:
CC: @marcinwrobel1986, @mperrinel

PR #435 should address the configure failures with Albany and Nalu reported in trilinos/Trilinos#9972 and trilinos/Trilinos#9973.

bartlettroscoe:
I am having to create reference builds of several application codes with Trilinos so that I can compare with the updated TriBITS build system.  This is taking a lot of time so I am putting this back in progress.

bartlettroscoe:
[...]

PR #435, opened 2021-12-09
Address <Package>Config.cmake file with includes to non-enabled subpackages (#433, #299)

bartlettroscoe:
This PR fixes the problem of TriBITS-generated `<Package>Config.cmake` files having includes for non-enabled subpackages.  This fixes the configure of Albany reported in trilinos/Trilihnos#9972 (and likely also Nalu in trilinos/Trilinos#9973).

This involved a significant update to how TribitsExampleProject package `depsX()` functions report their dependencies which required changing a lot of tests.  (The refactored code should hopefully make future updates easier.)

See the individual commits for more details.


bartlettroscoe:
CC: @marcinwrobel1986, @mperrinel

I will go ahead and merge and allows a post-merge review (and address any issues in a follow-up PR).

PR #441, opened 2022-01-27
TribitsExampleProject readme fix

marcinwrobel1986:
### This PR:
- updated README.md for TribitsExampleProject

### DIFF
--- a/tribits/examples/TribitsExampleProject/README.md
+++ b/tribits/examples/TribitsExampleProject/README.md
@@ -1,31 +1,35 @@
 # TribitsExampleProject Documentation
 
-The project TribitsExampleProject defines a TriBITS CMake project designed to
+The project `TribitsExampleProject` defines a TriBITS CMake project designed to
 provide a simple example to demonstrate how to use the TriBITS system to
 create a CMake build, test, and deployment system using a package-based
 architecture.
 
-To get started building you can create a directory structure like:
+To get started building you can create a directory structure with the source
+and build directories side-by-side like:
 
 ```
-    ~/TribitsExampleProject.base/
+    <some-base-dir>/
         TribitsExampleProject/   # This directory
         BUILDS/
           SERIAL/
 ```
 
-After that configure pointing to some value TriBITS source dir <tribits-dir>
-(i.e. TriBITS/tribits) with:
+For example, one could copy, clone, or symlink the `TribitsExampleProject` source
+dir (depending on where it came from) under `<some-base-dir>/`.
+
+First, configure by pointing to the TriBITS source dir `<tribits-dir>`
+(i.e. `TriBITS/tribits`) and the project source dir with:
 
 ```
-  cd BUILDS/SERIAL/ 
+  cd <some-base-dir>/BUILDS/SERIAL/ 
 
   cmake \
     -DTribitsExProj_TRIBITS_DIR=<tribits-dir> \
     -DTribitsExProj_ENABLE_TESTS=ON \
     -DTribitsExProj_ENABLE_ALL_PACKAGES=ON
     -DCMAKE_CXX_COMPILER=g++ \
-    ../..
+    ../../TribitsExampleProject
 ```
 
 and then build and test with:
--- a/tribits/examples/TribitsExampleProject/README.md
+++ b/tribits/examples/TribitsExampleProject/README.md
@@ -5,31 +5,22 @@ provide a simple example to demonstrate how to use the TriBITS system to
 create a CMake build, test, and deployment system using a package-based
 architecture.
 
-To get started building you can create a directory structure with the source
-and build directories side-by-side like:
+To build and test the project, one must first create a build directory and
+configure by pointing to the TriBITS source dir `<tribits-dir>`
+(i.e. `TriBITS/tribits`) and the `TribitsExampleProject` project source dir
+with:
 
 ```
-    <some-base-dir>/
-        TribitsExampleProject/   # This directory
-        BUILDS/
-          SERIAL/
-```
-
-For example, one could copy, clone, or symlink the `TribitsExampleProject` source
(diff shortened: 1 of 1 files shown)


tribits/examples/TribitsExampleProject/README.md
```diff
@@ -20,12 +20,12 @@ After that configure pointing to some value TriBITS source dir <tribits-dir>
 ```
   cd BUILDS/SERIAL/ 
 
-  cmake -D \
+  cmake \
     -DTribitsExProj_TRIBITS_DIR=<tribits-dir> \
     -DTribitsExProj_ENABLE_TESTS=ON \
     -DTribitsExProj_ENABLE_ALL_PACKAGES=ON
     -DCMAKE_CXX_COMPILER=g++ \
-    ../../examples/TribitsExampleProject
```
bartlettroscoe, line 28: Actually, based on the assumed directory structure:

```
    ~/TribitsExampleProject.base/
        TribitsExampleProject/   # This directory
        BUILDS/
          SERIAL/
```

we are both wrong.  The correct path for CMake is:

```
cmake \
<...> \
../../TribitsExampleProject
```

The directory `BUILDS` is not **under** the source dir  `TribitsExampleProject`, it is **beside** it.

bartlettroscoe (changes requested):
This was definingly a typo but the correct path based on the assumed directory structure is ../../TribitsExampleProject.

tribits/examples/TribitsExampleProject/README.md
marcinwrobel1986: Hello Ross,
I assumed: 
- one is currently in `TribitsExampleProject` (as it's marked # This directory) 
- one creates `BUILDS/SERIAL` dirs 
- then one (as it's given in the next step)  changes dir `cd BUILDS/SERIAL`
- and that's why `TribitsExampleProject` is in `../..`

marcinwrobel1986: But anyway please feel free to close or edit this as you think would be the best :)

bartlettroscoe: @marcinwrobel1986, can you take look at the updated file in this PR and provide a review and signoff?

bartlettroscoe: > provide a review and signoff?

@marcinwrobel1986 so it looks like you can provide a review comment of your own PR but can't approve it (of course).

tribits/examples/TribitsExampleProject/README.md
```diff
@@ -1,31 +1,35 @@
...
+  cd <some-base-dir>/BUILDS/SERIAL/ 
 
-  cmake -D \
+  cmake \
     -DTribitsExProj_TRIBITS_DIR=<tribits-dir> \
     -DTribitsExProj_ENABLE_TESTS=ON \
     -DTribitsExProj_ENABLE_ALL_PACKAGES=ON
     -DCMAKE_CXX_COMPILER=g++ \
-    ../../examples/TribitsExampleProject
+    ../../TribitsExampleProject
```
marcinwrobel1986, line 32: Hm.. From my point of view if one is in `cd <some-base-dir>/BUILDS/SERIAL/ ` then I would put here (in line 32) `<path-to-the-TribitsExampleProject>` or we could ask one to:
- first change directory to `TribitsExampleProject`
- make directories `BUILDS/SERIAL/ ` or just `build`
- and then leave here (in line 32) `../..` if `BUILDS/SERIAL/ ` or `..` if just `build`

bartlettroscoe: @marcinwrobel1986, I would generally advice not to put build directories under the source tree.  That is a bad practice in general, IMHO.  Because you can't just copy and move around git repos if you do that.  And Craig Scott, the author of "Professional CMake" agrees.  From the 10th edition in section 2.2 "Out-of-source Builds":

> This book will always use out-of-source builds and will follow the pattern of the source and build
directories being under a common parent.
>
> ...
>
> A variation on this used by some developers is to make the build directory a subdirectory of the
source directory. This offers most of the advantages of an out-of-source build, but it does still carry
with it some of the disadvantages of an in-source arrangement. Unless there is a good reason to
structure things that way, keeping the build directory completely outside of the source tree instead
is recommended.

For the life of me, I can't understand why people are putting the build dirs under the source dir.  That offers zero advantages that I can see.

marcinwrobel1986: Ok, then we have clear answer `<path-to-the-TribitsExampleProject>` - at least for me.
But IMHO as very often it depends on a usage, e.g. one has a PCBA, one downloads the software, builds it, flash it to the PCBA and forgets it exists until there is an update which needs to be flashed as well.

marcinwrobel1986: But in case we consider that could be true, but there is a question regarding a workflow. Maybe one works with IDE and it's easier when one has everything under project dir and just add `build` directory to `.gitignore`, but I agree that if one forgets about that and pushes build dir to the repository, then it could bring a mess.

bartlettroscoe: > Ok, then we have clear answer `<path-to-the-TribitsExampleProject>` - at least for me.

Yea, that is reasonable.  In that case, the instructions become even shorter.  Basically, you just create a build dir anywhere you want and then just point back to the source dir anywhere it exists.  And you just need to also point to a TriBITS implementation.

bartlettroscoe:
@marcinwrobel1986, please take a look at the updated README.md file.  Hopefully that will do the trick.

marcinwrobel1986 (commented):
All good:) agree, thanks!

bartlettroscoe (approved):
> All good:) agree, thanks!

@marcinwrobel1986, thanks for the feedback to improve this!

PR #445, opened 2022-02-03
Improve TribitsExampleProject and TribitsExampleApp readme

marcinwrobel1986:
- added `TribitsExProj_ENABLE_SECONDARY_TESTED_CODE=ON` as default
- added installation description
- added information regarding Fortran

PR #447, opened 2022-02-04
Adjust git checkout -B output regex for updated git version

bartlettroscoe:
The GitHub Actions VM env updated its git version (without warning) yesterday
and starting failing this test:

*  TriBITS_CTestDriver_TribitsExMetaProj_clone_default_branch_remote

This updated REGEX is not quite as precise but should work with both the older
and newer versions of git.


bartlettroscoe:
@marcinwrobel1986

FYI, this should fix the GHA test failures in the PRs like #445.  Once we confirm this fixes the test failures and it merges, the other PR tests should start passing.

Issue #443, opened 2022-02-01
Investigate and address issue with change to -isystem includes for IMPORTED targets (#299)


bartlettroscoe:
**CC:** @mperrinel, @marcinwrobel1986, @keitat 

**Parent Issues:**
* #299
* #433

**Other Linked issues**:

* Asking Kitware for help: https://gitlab.kitware.com/snl/project-1/-/issues/198

## Description of the problem

While studying a build errors reported with SPARC on the Vanguard system `van1-tx2` on the machine 'stria' using the ARM C++ 20.1 compiler (see [tribitsrefactoring#3](https://cee-gitlab.sandia.gov/rabartl/tribitsrefactoring/-/issues/3) and [SPAR-1074](https://sems-atlassian-srn.sandia.gov/browse/SPAR-1074)) for the updated TriBITS implementation increment to modern CMake targets in https://github.com/trilinos/Trilinos/pull/9978 (which is reapplying and fixing https://github.com/trilinos/Trilinos/pull/9894 after it was reverted), I traced the problem to a change in how include directories and specified when changing to modern cmake IMPORTED targets.

With old TriBITS, a downstream customer CMake project would get the Trilinos include directories by either setting the directory-level `INCLUDE_DIRECTORIES` property with:

```
include_directories( [SYSTEM] ${Trilinos_INCLUDE_DIRS})
```

or at the target level with:

```
target_include_directories(downstream PUBLIC|PRIVATE [SYSTEM] ${Trilinos_INCLUDE_DIRS})
```

and the project could either include these with `SYSTEM` or not and have them included on the compile line with `-isystem` or just `-I`, respectively.  SPARC and many other application codes usually omit `SYSTEM` so the Trilinos include dirs are specified as `-I` on the compile lines of the downstream customer object builds.

But with this change to modern CMake IMPORTED targets as per https://github.com/trilinos/Trilinos/pull/9894 (and then again after the revert of that with https://github.com/trilinos/Trilinos/pull/9978) to have the `INTERFACE_INCLUDE_DIRECTORIES` property set, by default, downstream targets add those include directories as `SYSTEM` includes and apply them as `-isystem` on the compile lines.  This changes the search order for header files and also silences warnings coming from those header files as noted in "Professional CMake: 10th edition" which states:

> It is also worth noting that paths specified by an imported target’s `INTERFACE_INCLUDE_DIRECTORIES`
property will be treated by consuming targets as though they were `SYSTEM` paths by default. 
[...]

PR #456, opened 2022-03-02
Add support for <Project>_IMPORTED_NO_SYSTEM and related (#443)

bartlettroscoe:
This PR add support for new option `<Project>_IMPORTED_NO_SYSTEM` that sets the property `IMPORTED_NO_SYSTEM` as per #443.

Please review the individual commits and the commit logs for the individual commits.

### How was this tested?

**Testing with TriBITS:**

There are strong native TriBITS tests to verify that this works as expected.  Ssee the new tests starting with `TriBITS_TribitsExampleApp_INCLUDE` for CMake 3.23+ shown [here](https://testing.sandia.gov/cdash/queryTests.php?project=TriBITS&filtercount=3&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=Linux-TRIBITS_MPI_DEBUG_CMAKE-3.23&field2=buildstarttime&compare2=81&value2=2022-02-24T18%3A43%3A21%20UTC&field3=testname&compare3=65&value3=TriBITS_TribitsExampleApp_INCLUDE) showing:

image

NOTE: Those tests only run with the CMake version is 3.23 or higher as per:

https://github.com/TriBITSPub/TriBITS/blob/bd2b1770c010d43783e3c3002ce1c0ddad06ddcf/test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake#L694-L698

and

https://github.com/TriBITSPub/TriBITS/blob/bd2b1770c010d43783e3c3002ce1c0ddad06ddcf/test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake#L703-L704

Therefore, we will need to wait until CMake 3.23 is released before we can turn on these tests in the GitHub Actions testing jobs.

**Testing with SPARC + Trilinos Reference build on 'van1-tx2':**

I also tested this with SPARC and confirmed that setting `-D Trilinos_IMPORTED_NO_SYSTEM=ON` with CMake 3.23-rc2 resulted in a successful build of SPARC on the 'van1-tx2' system 'stria' with the reference SPARC + Trilinos (see [internal issue tribitsrefactoring#3](https://cee-gitlab.sandia.gov/rabartl/tribitsrefactoring/-/issues/3#note_2203860))


bartlettroscoe:
@mperrinel, please review this PR, the individual commits and commit messages, and the updated documentation.  You can make suggesting by posting a stagged PR against this PR.

bartlettroscoe:
@mperrinel, I am going to add a new CHANGELOG.md entry for the addition of `<Project>_IMPORTED_NO_SYSTEM` variable and what it does to this PR branch.  That should take care of the new `<Project>_IMPORTED_NO_SYSTEM` variable.

bartlettroscoe:
@mperrinel and @marcinwrobel1986,

I will go ahead and merge this and then you can do a post-merge review.  Any issues can be addressed in a follow-up PR.

tribits/CHANGELOG.md
```diff
@@ -0,0 +1,328 @@
+----------------------------------------
+ChangeLog for TriBITS
+----------------------------------------
+
+## 2022-03-02:
+
+* **Added:** Added project-level cache varaible `<Project>_IMPORTED_NO_SYSTEM`
+  to set the `IMPORTED_NO_SYSTEM` property on the exported IMPORTED library
+  targets in the installed `<Package>Config.cmake` files (see updated TriBITS
+  documentation for `<Project>_IMPORTED_NO_SYSTEM`).
```
bartlettroscoe, line 10: Need to mention this requires CMake 3.23.0

tribits/doc/build_ref/TribitsBuildReferenceBody.rst
```diff
@@ -1401,6 +1401,26 @@ search for extra required libraries (such as the mpi library and the gfortran
...
+
+  -D ${PROJECT_NAME}_IMPORTED_NO_SYSTEM=ON
+
+is set, then all of the IMPORTED library targets exported into the set of
+installed ``<Package>Config.cmake`` files will have the ``IMPORTED_NO_SYSTEM``
+target property set.  This will cause downstream customer CMake projects to
+apply the include directories from these IMPORTED library targets as
+non-system include directories.  On most compilers, that means that the
+include directories will be listed on the compile lines with ``-I`` instead of
+with ``-isystem``.
```
bartlettroscoe, line 1421: Need to mention this requires CMake 3.23.0

bartlettroscoe (commented):
Need to mention IMPORTED_NO_SYSTEM requires CMake 3.23.0

test/core/ExamplesUnitTests/TribitsExampleProject_Tests.cmake
```diff
@@ -2801,12 +2801,13 @@ tribits_add_advanced_test( TribitsExampleProject_compiler_flags
   OVERALL_WORKING_DIRECTORY  TEST_NAME
   OVERALL_NUM_MPI_PROCS  1
   EXCLUDE_IF_NOT_TRUE  IS_REAL_LINUX_SYSTEM  COMPILER_IS_GNU
-    ${PROJECT_NAME}_ENABLE_Fortran
+    ${PROJECT_NAME}_ENABLE_Fortran  NINJA_EXE
```
mperrinel, line 2804: You choose to use Ninja to get verbose output with cmake 3.23. Do we have the also the verbose output with Ninja and cmake < 3.23 ?

tribits/core/package_arch/TribitsLibraryMacros.cmake
```diff
@@ -988,12 +988,14 @@ function(tribits_add_library LIBRARY_NAME_IN)
...
 
-    #
     # Add ALIAS library <PackageName>::<libname>
-    #
-
     add_library(${PACKAGE_NAME}::${LIBRARY_NAME} ALIAS ${LIBRARY_NAME})
 
+    # Optionally Set IMPORTED_NO_SYSTEM
+    if (${PROJECT_NAME}_IMPORTED_NO_SYSTEM)
+      set_target_properties(${LIBRARY_NAME} PROPERTIES IMPORTED_NO_SYSTEM TRUE)
```
mperrinel, line 996: Do we need to do the contrary in case of `if (! ${PROJECT_NAME}_IMPORTED_NO_SYSTEM)` ? I don't think since it is the default value. But that could also be a way to ensure that we know exactly the value of this property.

tribits/doc/build_ref/TribitsBuildReferenceBody.rst
bartlettroscoe: Fixed in PR #459

tribits/CHANGELOG.md
bartlettroscoe: Fixed in PR #459.

bartlettroscoe:
> Therefore, we will need to wait until CMake 3.23 is released before we can turn on these tests in the GitHub Actions testing jobs.

Actually, I was able to turn on these tests in GitHub Actions testing in PR #458.

tribits/core/package_arch/TribitsLibraryMacros.cmake
bartlettroscoe: No, I think it is better to leave a property at its default value if you don't set it to a non-default value that would change behavior.  Otherwise, we would be setting all kinds of properties to their default value in other such cases which would just confuse developers trying to inspect the set properties and debug things.  There is no recommendation to set properties to their default values in "Professional CMake: 10th Edition" (see Section "9.8. Recommended Practices" in "Chapter 9. Properties").

test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake
```diff
@@ -647,6 +647,151 @@ TribitsExampleApp_ALL_ST(ByPackage STATIC)
...
+  elseif (byProjectOrPackage STREQUAL "ByPackage")
+    set(findByProjectOrPackageArg -DTribitsExApp_FIND_INDIVIDUAL_PACKAGES=ON)
+    set(foundProjectOrPackageStr "Found SimpleCxx")
+  else()
+    message(FATAL_ERROR "Invaid value for findByProjectOrPackageArg='${findByProjectOrPackageArg}'!")
+  endif()
+
+  TribitsExampleApp_process_sharedOrStatic_arg()
+
+  if (importedNoSystem STREQUAL "IMPORTED_NO_SYSTEM")
```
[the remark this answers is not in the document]
bartlettroscoe: >  I would use a bool instead of a string

@mperrinel, I strongly disagree.

Bools are almost always a bad choice, except for one-arg functions and some local variables where the meeting is obvious.  And bools should be avoided as arguments in multi-argument functions.  See section "12.6 Enumerated Types" in "Code Complete: 2nd Edition" which says:

> Enumerated types are especially useful for defining routine parameters

and then gives an example of a terrible function that takes multiple bools like:

```
int result = RetrievePayrollData(data, true, false, false, true)
```

There is similar advice in the book "Clean Code":

> Flag arguments are ugly.  Passing a boolean into a function is truly a terrible practices.

From the same author [Robert C. Martin's Clean Code Tip #12: Eliminate Boolean Arguments](https://www.informit.com/articles/article.aspx?p=1392524).

And then there is:

* https://alexkondov.com/should-you-pass-boolean-to-functions/
* https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/
* https://medium.com/swlh/solving-the-boolean-identity-crisis-33eeccde2c96

@mperrinel, where do you find guidance/best practice to have multiple bool arguments passed to a function?  What books have you read that have defined your software development practicies?

test/core/ExamplesUnitTests/TribitsExampleProject_Tests.cmake
bartlettroscoe: I have no idea but it does not really matter in this context.

test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake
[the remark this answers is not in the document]
mperrinel: Hi Ross,
Yes I understand your point and I read all the documentation you sent, [except the book of Robert C. Martin’s Clean Code Tip #12: Eliminate Boolean Arguments]. My point with the Boolean suggestion was in order to simplify the code for more robustness :
```
if (byProjectOrPackage STREQUAL "ByProject")
  set(findByProjectOrPackageArg -DTribitsExApp_FIND_INDIVIDUAL_PACKAGES=OFF)
  set(foundProjectOrPackageStr "Found TribitsExProj")
elseif (byProjectOrPackage STREQUAL "ByPackage")
  set(findByProjectOrPackageArg -DTribitsExApp_FIND_INDIVIDUAL_PACKAGES=ON)
  set(foundProjectOrPackageStr "Found SimpleCxx")
else()
  message(FATAL_ERROR "Invaid value for findByProjectOrPackageArg='${findByProjectOrPackageArg}'!")
endif()
```
which would have become :
```
if (isPackaged)
  set(findByProjectOrPackageArg -DTribitsExApp_FIND_INDIVIDUAL_PACKAGES=ON)
  set(foundProjectOrPackageStr "Found SimpleCxx")
else()
  set(findByProjectOrPackageArg -DTribitsExApp_FIND_INDIVIDUAL_PACKAGES=OFF)
  set(foundProjectOrPackageStr "Found TribitsExProj")
endif()
```
 In the current implementation, it is possible to make a mistake when calling the function, which will result in entering the else path. If using a Boolean instead, this mistake is no longer possible (it is a package or it is not a package).
Also, I don’t think that is it is necessarily always the best practice to hard code string values directly into a function. Indeed, if one creates another function that uses the **byProjectOrPackage** enum, one will then have to test again with the hard coded string, hereby risking to introduce typos. In addition, if one needs to change it later, that will result in more work ass all the functions using the hard coded string will have to be modified as well. Moreover, users of the function can easily introduce typos when they calls the function, which is simply impossible with the Boolean.
I understand that, with a Boolean, calling the method that way : `TribitsExampleApp_INCLUDE(TRUE SHARED IMPORTED_NO_SYSTEM)` is not good because one cannot what **TRUE** could mean in this context, and one needs to look into the function in order to understand it. However, I think that could be easily fixed by creating a variable as I suggested like that:` set(isPackage TRUE)`.
Finally, I did not realize that **byProjectOrPackage** was an enum because it seems to have only two possible values and the function uses it values to define the same CMake option **TribitsExApp_FIND_INDIVIDUAL_PACKAGES**, which contains **PACKAGES** in its name. This appeared to me as being more associated with the **ByPackage** enum value. That is why I thought that using a Boolean based on the **ByPackage** enum value named **isPackage** would be a good idea.

tribits/core/package_arch/TribitsLibraryMacros.cmake
mperrinel: Ok, I understand

test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake
```diff
@@ -647,6 +647,151 @@ TribitsExampleApp_ALL_ST(ByPackage STATIC)
 TribitsExampleApp_ALL_ST(ByPackage SHARED)
 
 
+################################################################################
+
+
+function(TribitsExampleApp_INCLUDE byProjectOrPackage sharedOrStatic importedNoSystem)
```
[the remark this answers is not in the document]
mperrinel: I understand. Since it is a enum, we should find a simple name that matches all the values. What do you think about `ConfigCMakeType` ?

test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake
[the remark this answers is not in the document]
bartlettroscoe: > My point with the Boolean suggestion was in order to simplify the code for more robustness

So an enum has to have more than two values or you should always use a bool?  I strongly disagree with that.

Let's agree to disagree on this one

bartlettroscoe:
This review is complete.  Moving to "Done"

test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake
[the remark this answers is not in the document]
bartlettroscoe: Let's discuss this briefly at the next meeting to explain the massive advantages of passing enum-like strings in this case vs. bool arguments.

But basically, the huge advantage of using enum-like string args like this is that test definition calls like:

https://github.com/TriBITSPub/TriBITS/blob/a72c0d5266eec342751abb7b1e474706ef5b9eb2/test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake#L390-L392

showing:

```
TribitsExampleApp_NoFortran(FULL STATIC)
TribitsExampleApp_NoFortran(COMPONENTS STATIC)
TribitsExampleApp_NoFortran(COMPONENTS SHARED)
```

exactly map to the actual CTest names like shown at [here](https://testing.sandia.gov/cdash/build/10696211/configure) showing:

```
-- TriBITS_TribitsExampleApp_NoFortran_FULL_STATIC: Added test (BASIC, PROCESSORS=1)!
-- TriBITS_TribitsExampleApp_NoFortran_COMPONENTS_STATIC: Added test (BASIC, PROCESSORS=1)!
-- TriBITS_TribitsExampleApp_NoFortran_COMPONENTS_SHARED: Added test (BASIC, PROCESSORS=1)!
```

And we see the same things with the other tests like at:

* https://github.com/TriBITSPub/TriBITS/blob/master/test/core/ExamplesUnitTests/TribitsExampleApp_Tests.cmake#L786-L790

showing:

```
TribitsExampleApp_INCLUDE(ByProject STATIC "")
TribitsExampleApp_INCLUDE(ByProject SHARED IMPORTED_NO_SYSTEM)
TribitsExampleApp_INCLUDE(ByProject STATIC IMPORTED_NO_SYSTEM)
TribitsExampleApp_INCLUDE(ByPackage SHARED IMPORTED_NO_SYSTEM)
TribitsExampleApp_INCLUDE(ByPackage STATIC IMPORTED_NO_SYSTEM)
```

and the associated tests created shown [here](https://testing.sandia.gov/cdash/build/10696211/configure) showing:

```
-- TriBITS_TribitsExampleApp_INCLUDE_ByProject_STATIC: Added test (BASIC, PROCESSORS=1)!
-- TriBITS_TribitsExampleApp_INCLUDE_ByProject_SHARED_IMPORTED_NO_SYSTEM: Added test (BASIC, PROCESSORS=1)!
-- TriBITS_TribitsExampleApp_INCLUDE_ByProject_STATIC_IMPORTED_NO_SYSTEM: Added test (BASIC, PROCESSORS=1)!
-- TriBITS_TribitsExampleApp_INCLUDE_ByPackage_SHARED_IMPORTED_NO_SYSTEM: Added test (BASIC, PROCESSORS=1)!
-- TriBITS_TribitsExampleApp_INCLUDE_ByPackage_STATIC_IMPORTED_NO_SYSTEM: Added test (BASIC, PROCESSORS=1)!
```

It is pretty much impossible to have more self-documenting test code than that.

Compare that to switching to bool args as:

```
TribitsExampleApp_INCLUDE(False False False)
TribitsExampleApp_INCLUDE(False True True)
TribitsExampleApp_INCLUDE(False False True)
TribitsExampleApp_INCLUDE(True True True)
TribitsExampleApp_INCLUDE(True False True)
```

What tests to those calls correspond to? What are they actually testing?

Can you see how terrible that is?

rabartlett1972 (commented):
LGTM

PR #474, opened 2022-03-29
TribitsExample[Project2|App2]: Add support for optional package and TPL dependencies (#299)

bartlettroscoe:
Updated `TribitsExampleProject2` and `TribitsExampleApp2` for optional dependencies being disabled.  This required adding the generation of configured `<Package>_config.h` files and adding `<Package>_ENABLE_<dep>` vars to the `<Package>Config.cmake` files.  (This shows how we can export info about packages to their `<Package>Config.cmake`  files so that downstream packages can get access to that info.  There is other types of info we will need to do this for as well as we allow building and installing TriBITS packages (like in Trilinos) individually.

And with this PR in place, now I can add support for TPL dependencies!

This part of #299.


test/core/ExamplesUnitTests/TribitsExampleProject2_Tests.cmake
```diff
@@ -264,6 +286,130 @@ TribitsExampleProject2_find_tpl_parts(SHARED  CMAKE_PREFIX_PATH_ENV)
...
+
+  TribitsExampleProject2_test_setup_header()
+
+  set(tplInstallBaseDir
+    "${TribitsExampleProject2_Tpls_install_${sharedOrStatic}_DIR}")
+
+  set(testNameSuffix "")
+  set(tplLibAndIncDirsArgs "")
+  set(cmakePrefixPathCacheArg "")
+  set(cmakePrefixPathEnvArg "")
```
rabartlett1972, line 299: This variable is not needed in this function.  Should remove.

rabartlett1972 (changes requested):
I don't see any major issues, just a var to clean up.

test/core/ExamplesUnitTests/TribitsExampleProject2_Tests.cmake
bartlettroscoe: All of these unneeded temp vars are removed in new commit https://github.com/TriBITSPub/TriBITS/pull/474/commits/7b552d36100f5b6851cd49493eb3a5617c8c3bb1.

rabartlett1972 (approved):
Looks good.

bartlettroscoe:
@keitat, you can review some if you want.

PR #479, opened 2022-04-29
Switch over fully to modern CMake targets and updates for testing against Trilinos (#299)

bartlettroscoe:
This PR contain the final changes to switch over the only modern CMake targets `<Package>::all_libs` for all internal packages and external packages/TPLs.  This strips out all of the old TriBITS logic that manually handled the list of include dirs and libraries.  Modern CMake targets are now used for handling off of that.  And as you can see, more lines of code were deleted than added in this changeset.

The updated logic in `tribits_add_library()` and `tribits_add_executable()` is now pretty easy to follow.  There is no more complex magic.

This PR also contain any changes needed for building Trilinos 'develop' with this version of TriBITS.  (I just copied over the FindTPLNetcdf.cmake file from Trilinos).

This PR, however, does contain more backward compatible breaking changes (see the updates to the `CHANGELOG.md` file).

Please see the individual commit log messages for more details.

## Review Checklist

* [x] Read over updated items in [`CHANGLOG.md`](https://github.com/TriBITSPub/TriBITS/pull/479/files#diff-f8f03bafafec69deb3fdf43f40078eaf363653cc6b341088718a572138f048e9)
* [x] Read over updated text in [`TribitsGuidesBody.rst`](https://github.com/TriBITSPub/TriBITS/pull/479/files#diff-da435984138267dc9947fab5adf5227aa2e13ca792d1905d88e8e689d797f472)

PR #480, opened 2022-05-03
Add flat data-structures to loop over direct package dependencies (#63, #299)

bartlettroscoe:
This adds new flat list vars `<Package>_[LIB|TEST]_[ENABLED|ALL]_DEPENDENCIES` to make it easy to loop over all of the upstream internal and external packages/TPLs.  This will make it easier to fully switch to new modern cmake targets `<Package>::all_libs` (#299).  It will also make it easier to finish merging handling and support for TriBITS internal packages and TPLs (#63).   It also adds setting the vars `<Package>_ENABLE_<depPkg>=ON` for requires direct dependencies and tricky logic related to required subpackages and their parent package.

This is a bit of a large PR in terms of lines of code effected and it impacts several tests, mostly just adding additional checks to existing tests.  There is also contains some other refactorings.  There are also some other small fixes and refactorings that I noticed while doing this work.

The next PR after this will be the actual change over to all linking and include directory management through modern `<Package>::all_libs` targets.

## Review Checklist

* [ ] Read over the individual commit messages and the patches in those commits
* [ ] Focus on the documentation updates for understandability and correctness
* [ ] Look over production code changes under `tribits/`
* [ ] Just skim changes/additions to the tests under `test/`


rabartlett1972 (approved):
I will allow this to merge to 'master' and we can do a post-merge PR if needed.  We need to keep moving since I need to merge PR #479 based on these changes soon.

bartlettroscoe:
FYI: Note that the decision to set `<Package>_ENABLE_<depPkg>` to `ON` even for required upstream internal and external packages in commit https://github.com/TriBITSPub/TriBITS/pull/480/commits/175ea956318f1789cdc452715cead1213fa131e4 was the right decision but it technically breaks backward compatibility and it actually broke the Trilinos build as described in:

* https://github.com/trilinos/Trilinos/issues/10534

But the reason that it broke the build is that someone changed a dependency from optional to required, but did not remove the conditional behavior based on `<Package>_ENABLE_<depPkg>` (which was now no longer being set to `ON`).  In that case, `Isorropia_ENABLE_EpetraExt` went from being set to `ON` when its dependence on `EpetraExt` was optional and then become undefined (and therefore implicitly off) when its dependence on `EpetraExt` become required.  As a result, a bunch of tests got disabled 10+ years ago.  And when Trilinos was tested against the new version of TriBITS, suddenly `Isorropia_ENABLE_EpetraExt` was set to `ON` again and code that had not been built in over 10 years was no longer building.

But if the original version of TriBITS has been setting `<Package>_ENABLE_<depPkg>` to `ON` for required dependencies, then that test code in Isorropia would have still be enabled and would have kept building and running for the last 10 years.

So to make things more robust when a package dependency goes from optional to required (like what happened as described in https://github.com/trilinos/Trilinos/issues/10534), we want to set `<Package>_ENABLE_<depPkg>` to `ON` even for required dependencies.

NOTE: I have documented this break in backward compatibility in the follow-on PR #479 in commit https://github.com/TriBITSPub/TriBITS/pull/479/commits/630d877449b842c557babc0a2c08bb6db57d9656.

PR #486, opened 2022-06-10
Address review feedback from PR #485

bartlettroscoe:
This PR addresses post-merge review comments from PR #485 from @KyleFromKitware.

See the commits and commit logs for more details.

PR #487, opened 2022-06-10
Remove line containing '#' before all @MACRO and @FUNCTION blocks

bartlettroscoe:
I did this as part of addressing post-merge review comments for PR #485 but I did not want to include this glut of files in the follow-up PR #486.


rabartlett1972 (approved):
Trivial changes

PR #488, opened 2022-06-17
Fix some spelling errors

gsjaardema:
Just fixing a few random spelling errors.  Nothing that affects the execution although there is one that appears in an error/warning message visible to users.


bartlettroscoe (approved):
LGTM.  Thanks @gsjaardema!

I don't think this will break any TriBITS tests or any user code.  This only seems to impact the doc string for a single cache var.

PR #490, opened 2022-06-28
Call tribits_tpl_find_include_dirs_and_libraries(CUDA) (#299)

bartlettroscoe:
Every TPL, envy dummy TPLs like CUDA, need to define the <tplName>::all_libs target.  Trilinos needs this for a CUDA build.

This passes the Trilinos CUDA ATDM build on vortex and seems to be allowing the build to continue.

There are no native TriBITS tests that test this and no way to test it (unless we add an example TriBITS package that depends on CUDA and then we set up a CUDA PR build, which is not possible with GitHub Actions).


bartlettroscoe:
I was able to build and test Kokkos successfully under Trilinos.  Therefore, I will merge this.

PR #491, opened 2022-06-28
Remove assert for TPL_<tplName>_LIBRARIES (#299)

bartlettroscoe:
With the new TriBITS TPl implementation, the is no need for a FindTPL<tplName>.cmake file to set the var TPL_<tplName>_LIBRARIES so there is no need to assert it.

This came up when refactoring the FindTPLCUBLAS.cmake, FindTPLCUSPARSE.cmake and FindTPLCUSOLVE.cmake files in Trilinos as part of the PR trilinos/Trilinos#10614.


bartlettroscoe:
The basic build of Trilinos packages worked and the TriBITS test suite in the GHA builds passed, so I will merge this.

NOTE: This is just a single line of code being removed here so this does not require a review, IMHO.

PR #504, opened 2022-07-26
Change type to IMPORTED STATIC for *.a file (trilinos/Trilinos#10774)

bartlettroscoe:
Resolves link problems on ATS1 described in https://github.com/trilinos/Trilinos/issues/10774#issuecomment-1194856254 and https://github.com/trilinos/Trilinos/issues/10774#issuecomment-1192075276.  See commit messages for more details.

NOTE: This is not really a TriBITS bug.  I believe this to be a CMake bug that we are working around here.  And this makes this TriBITS code just a little bit less general and perhaps more fragile.

### DIFF
--- a/test/core/TribitsExternalPackageWriteConfigFile_UnitTests.cmake
+++ b/test/core/TribitsExternalPackageWriteConfigFile_UnitTests.cmake
@@ -619,7 +619,7 @@ add_library(SomeTpl::lib1 IMPORTED INTERFACE GLOBAL)
 set_target_properties(SomeTpl::lib1 PROPERTIES
   IMPORTED_LIBNAME "lib1")
 
-add_library(SomeTpl::somelib IMPORTED UNKNOWN GLOBAL)
+add_library(SomeTpl::somelib IMPORTED STATIC GLOBAL)
 set_target_properties(SomeTpl::somelib PROPERTIES
   IMPORTED_LOCATION "/some/other/path/to/libsomelib.a")
 target_link_libraries(SomeTpl::somelib
@@ -676,7 +676,7 @@ add_library(SomeTpl::lib1 IMPORTED INTERFACE GLOBAL)
 set_target_properties(SomeTpl::lib1 PROPERTIES
   IMPORTED_LIBNAME "lib1")
 
-add_library(SomeTpl::somelib IMPORTED UNKNOWN GLOBAL)
+add_library(SomeTpl::somelib IMPORTED STATIC GLOBAL)
 set_target_properties(SomeTpl::somelib PROPERTIES
   IMPORTED_LOCATION "/some/other/path/to/libsomelib.a")
 target_link_libraries(SomeTpl::somelib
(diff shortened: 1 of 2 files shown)


rabartlett1972 (approved):
Urgent to get this merged so we will do a post-merge review

bartlettroscoe:
Hello @KyleFromKitware, can you please do a post-merge review of this PR?  Also, can you comment on the behavior of CMake with IMPORTED UNKNOWN targets with `*.a` files adding `-Wl,-Bdynamic` described in https://github.com/trilinos/Trilinos/issues/10774#issuecomment-1194856254?  Is this a defect in CMake?  (And if it is not considered a defect, then what is the logic for that behavior?)

KyleFromKitware:
I'm not sure exactly why `-Wl,-Bdynamic` is being added. My best guess is that when the dependency type is `UNKNOWN`, CMake is taking the approach of "better to be safe than sorry" in case the dependency is a shared library, but forgetting that `-Wl,-Bdynamic` not even needed for linking static libraries.

The changes themselves look fine.

KyleFromKitware:
> Can we post a CMake Issue for this to try to this changed in future versions of CMake or at least have a broader conversation about this to document this issue?

Yes, go ahead and open an issue.

bartlettroscoe:
CC: @KyleFromKitware, @bradking 

So let's get really specific about this, why would cmake ever put a `*.a` file on the link line prefixed with `-Wl,-Bdynamic`?  Is there any use case where that makes sense?  My understanding is that the purpose for the option  `-Wl,-Bdynamic` to be used with `-l<libname>` so that it tells the compiler to find a shared library instead of a static library.  But the options `-Wl,-Bdynamic` and `Wl,-Bstatic` may no sense when passing in a full library path file (which the compiler can figure out for its own if it is static or shared).

So how can this not be a CMake defect that it would generate the link options:

```
-Wl,-Bdynamic /some/path/to/libstaticlib.a
```

under any situation, unless the user did something stupid and generated the IMPORTED target with SHARED as with:

```
add_library(<prefix>::<libname> IMPORTED SHARED GLOBAL)
set_target_properties(<prefix>::<libname>  PROPERTIES
  IMPORTED_LOCATION "/some/path/to/libstaticlib.a")
```

?

KyleFromKitware:
Yeah, I'm guessing this was a simple oversight on CMake's part. Please open a CMake issue.

bartlettroscoe:
> Yeah, I'm guessing this was a simple oversight on CMake's part. Please open a CMake issue.

I created the CMake issue:

* "CMake incorrectly prefixes *.a on link line with -Wl,-Bdynamic for IMPORTED UNKNOWN targets": https://gitlab.kitware.com/cmake/cmake/-/issues/23777

bartlettroscoe:
Given the issue:

* "CMake incorrectly prefixes *.a on link line with -Wl,-Bdynamic for IMPORTED UNKNOWN targets": https://gitlab.kitware.com/cmake/cmake/-/issues/23777

has been created and @KyleFromKitware reviewed and approved this PR with no issues, I will move to "Done".

Issue #505, opened 2022-07-27
Fix find_package(<Package>) calls from TriBITS-generated <Package>Config.cmake files from different dirs (#299)


bartlettroscoe:
It was reported that the new TriBITS-generated `<Pacakge>Config.cmake` files don't work correctly when calling `find_package(<Package>)` from different directories for different packages that have a dependency relationship.  This was reported from a SEACAS project customer that is trying to call `find_package()` from two different directories for two different SEACAS subpackages from that have a dependency.  An small reproducer was provided:

* [cmake-seacas-error.tar.txt](https://github.com/TriBITSPub/TriBITS/files/9203228/cmake-seacas-error.tar.txt)

This has the directory structure:

```
 ~/tree.py -cf .
  cmake-seacas-error/
  +-CMakeLists.txt
  +-a/
  | +-CMakeLists.txt
  +-b/
  | +-CMakeLists.txt
  | +-b.cpp
```

The file `cmake-seacas-error/a/CMakeLists.txt` has:

```
find_package(SEACASExodus REQUIRED HINTS /projects/seacas/linux_rhel7/current)
```

while the file `cmake-seacas-error/b/CMakeLists.txt` has:

```
find_package(SEACASNemesis REQUIRED HINTS /projects/seacas/linux_rhel7/current)
```

It is reported that the second call to `find_package(SEACASNemesis ...)` results in the error:

```
08:13:39    Found package configuration file:
08:13:39  
08:13:39      /projects/seacas/linux_rhel7/current/lib/cmake/SEACASNemesis/SEACASNemesisConfig.cmake
08:13:39  
08:13:39    but it set SEACASNemesis_FOUND to FALSE so package "SEACASNemesis" is
08:13:39    considered to be NOT FOUND.  Reason given by package:
08:13:39  
08:13:39    The following imported targets are referenced, but are missing:
08:13:39    SEACASExodus::all_libs
```

This suggests a problem with how the new TriBITS-generated files are written using `include_guard(GLOBAL)`.

bartlettroscoe:
I was able to reproduce the error on 'hpws055' using the slightly simplified project:

```
$ ~/tree.py -cf cmake-seacas-error2
  cmake-seacas-error2/
  +-CMakeLists.txt
  +-a/
  | +-CMakeLists.txt
  +-b/
  | +-CMakeLists.txt
```

with the project:

[...]

PR #506, opened 2022-07-28
Fix find_package() from separate subdirs for <Package>Config.cmake files (#505)

bartlettroscoe:
This should fix the use case described in #505.   I produced a TriBITS test that exposed the problem and updated the code to fix it.  I think what is here now works as find_package() should from different subdirs for dependent packages but still avoid the scalability problem described in https://gitlab.kitware.com/cmake/cmake/-/issues/23685.  This also updates the `<tplName>Config.cmake` files to act in a consistent way with CMake-generated imported targets (using `export()` command).

See the individual commit log messages for details.


rabartlett1972 (approved):
I will approve without an initial review so this can be updated to SEACAS and address but #505.

Then we can do a post-merge review.

bartlettroscoe:
Hello @KyleFromKitware, can you please do a post-merge review of this PR?  Thanks!

KyleFromKitware:
LGTM +1

PR #507, opened 2022-08-06
Change from 'http' to 'https' for CDash submissions

bartlettroscoe:
See commit log msg for details.  This is not a bug in TriBITS, just a change to the CDash site.


rabartlett1972 (approved):
If it passes the GitHub Actions builds, which includes submitting to testing-dev.sandia.gov/cdash/ then it means it works and it can merge.

rabartlett1972 (approved):
Re-approve after rebase

Issue #508, opened 2022-08-09
Switch from testing.sandia.gov to my.cdash.org


bartlettroscoe:
### SNL Kitware Issue

* [snl-kitware#213](https://gitlab.kitware.com/snl/project-1/-/issues/213)

### Description

It seems that testing.sandia.gov/cdash/ is going away (see https://github.com/trilinos/Trilinos/issues/10852), I need to start posting TriBITS results to my.cdash.org.

bartlettroscoe:
PR #509 was merged.  I had not created the "Pull Request" group yet and the PR results when to "Experimental".  I will see if the next PR correctly posts results to the "Pull Request" group before I close this issue.

bartlettroscoe:
FYI: After this change, the results for the PR #512 shown [here](https://github.com/TriBITSPub/TriBITS/actions/runs/2842857588) are not submitting to the TriBITS project on my.cdash.org.   Therefore, I created the PR #513 to switch back.

bartlettroscoe:
@zackgalbreath, I would like to give this switch another shot to see if we can make this work.  Then we would not have to deal with trying to keep an external CDash server under a sandia.gov domain.  I have already created the project:

* https://my.cdash.org/index.php?project=TriBITS

But I could  never get the submits to the "Pull Request" group working correctly.  Let me set up a local configure and driver scenario so you can reproduce the attempts on your side.

bartlettroscoe:
I posted the new PR #517 to see if the submits go to:

* https://my.cdash.org/index.php?project=TriBITS&filtercount=1&showfilters=1&field1=groupname&compare1=61&value1=Pull%20Request

correctly.

bartlettroscoe:
Since there was good results from posting PR #517, I decided to merge this.  This goes back to posting to my.cdash.org and so far so good.

bartlettroscoe:
I merged PR #519.  Now I put this in review for a few day days to make sure that "Nightly", "Continuous" and "Pull Request" build submit correctly.

bartlettroscoe:
Okay, the "Nightly" GitHub Actions builds showed up today at:

* https://my.cdash.org/index.php?project=TriBITS&date=2022-08-19&filtercount=0&showfilters=1

and

* https://testing-dev.sandia.gov/cdash/index.php?project=TriBITS&date=2022-08-19&filtercount=0&showfilters=1

[...]

PR #509, opened 2022-08-09
Change to post to my.cdash.org (#508)

bartlettroscoe:
I just created that project on my.cdash.org.

Addresses #508.


rabartlett1972 (approved):
Let's try posting to my.cdash.org for a while and see what happens.

PR #512, opened 2022-08-11
Fix test dependencies for subpackages and subpackage tests/examples enabless (#63, #268, #299, #510)

bartlettroscoe:
Addresses specifically the defect in #510 which resulted from work done as part of #63 and #299.  This also mostly addresses the scope of #268.

Once this is updated to Trilinos, it will fix https://github.com/trilinos/Trilinos/issues/10842.

See the specific git commit log messages for more detail.

PR #515, opened 2022-08-13
Replace leading tabs with 8 spaces (#512)

bartlettroscoe:
Somehow tabs got inserted instead of spaces when indenting in these files.

I replaces tabs with 8 spaces and things seem to look about correct.


rabartlett1972 (dismissed):
If it passes the tests, this is good to merge.

rabartlett1972 (approved):
Reapprove after forced push

PR #530, opened 2022-10-14
Set new package deps graph vars combining external and internal packages (#63)

bartlettroscoe:
## Description

This PR sets up the new package dependency vars described in the section [[Variables defining the package dependencies graph](https://tribitspub.github.io/TriBITS/maintainers_guide/index.html#id227)](https://tribitspub.github.io/TriBITS/maintainers_guide/index.html#variables-defining-the-package-dependencies-graph) (except the list names `<Package>_FORWRD_[LIB|TEST]_DEP_PACKAGES` was changed to `<Package>_FORWRD_[LIB|TEST]_DEFINED_DEPENDENCIES` for consistency with the new vars and to contrast with the old vars).  This is another increment working to complete #63.

See the individual commits for more details.

## Instructions for Reviewers

* It may be easier to review this PR one commit at a time (note next item)
* NOTE: There was a change made in the handling of `<Project>_ASSERT_DEFINED_DEPENDENCIES` and `<Project>_ASSERT_MISSING_PACKAGES` part way through the set of commits.  (Therefore, be sure to compare with the final version of the file before putting any time into commenting an individual git commit patch.)


bartlettroscoe:
Hello @KyleFromKitware, can you please do a full review of this PR and select "Approve" if you approve (or suggest changes)?  Otherwise, I will be working on the next PR based in this PR's branch so there is no huge rush to get this merged.

test/core/DependencyUnitTests/CMakeLists.txt
```diff
@@ -547,7 +547,8 @@ create_dependency_handling_test_case(
   ARGS
     -DTrilinos_ENABLE_ALL_PACKAGES:BOOL=ON
     -DTrilinos_DUMP_PACKAGE_DEPENDENCIES:BOOL=ON
--DTrilinos_WARN_ABOUT_MISSING_EXTERNAL_PACKAGES=ON
+    -DTrilinos_DUMP_FORWARD_PACKAGE_DEPENDENCIES:BOOL=ON
+    -DTrilinos_WARN_ABOUT_MISSING_EXTERNAL_PACKAGES=ON
```
KyleFromKitware, line 551: **Question:** Should there be a `:BOOL` here?

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
```diff
@@ -541,7 +545,7 @@ endmacro()
 #
 # See `Function call tree for constructing package dependency graph`_.
 #
-function(tribits_set_dep_packages  packageName  testOrLib  requiredOrOptional  pkgsOrTpls)
+macro(tribits_set_dep_packages  packageName  testOrLib  requiredOrOptional  pkgsOrTpls)
```
KyleFromKitware, line 548: Be careful changing a `function` to a `macro` - this will leak every single variable used inside the macro. It might be better to `set(... PARENT_SCOPE)` on the variables you need to return instead.

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
```diff
@@ -616,29 +629,32 @@ macro(tribits_set_dep_packages__handle_undefined_pkg  packageName  depPkg
           " is missing and ${depPkg}_ALLOW_MISSING_EXTERNAL_PACKAGE ="
           " ${${depPkg}_ALLOW_MISSING_EXTERNAL_PACKAGE}!")
       endif()
-      if (requiredOrOptional STREQUAL "REQUIRED")
+      if ("${requiredOrOptional}" STREQUAL "REQUIRED")
```
KyleFromKitware, line 632: **Question:** Any particular reason for this change? The way you had it before is the canonical way of comparing a variable to some value (though the quoted method works too of course.)

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
```diff
@@ -660,13 +676,19 @@ endmacro()
 #
 # See `Function call tree for constructing package dependency graph`_.
 #
-function(tribits_append_forward_dep_packages  packageName  inputListType)
+macro(tribits_append_forward_dep_packages  packageName  libOrTest  requiredOrOptional)
```
KyleFromKitware, line 679: Again, careful with `macro`, might be better to use `set(... PARENT_SCOPE)`.

tribits/core/utils/TribitsAddEnumCacheVar.cmake
```diff
@@ -0,0 +1,89 @@
...
+  tribits_assert_parse_arg_one_or_more_values(PREFIX ${argMultiValArgKeywords}) 
+  # Human readable list of allowed values: '<val0>', '<val1>', ...
+  string(REPLACE ";" "', '" validStringValuesListStr "'${PREFIX_ALLOWED_STRINGS_LIST}'")
+  # Set cache var
+  set(${cacheVarName}  ${PREFIX_DEFAULT_VAL}  CACHE  STRING
+    "${PREFIX_DOC_STRING}.  Valid values: ${validStringValuesListStr} (default '${PREFIX_DEFAULT_VAL}')")
+  if (PREFIX_IS_ADVANCED)
+    mark_as_advanced(${cacheVarName})
+  endif()
+  set_property(CACHE  ${cacheVarName}  PROPERTY  STRINGS
```
KyleFromKitware, line 81: Good use of `STRINGS` here. This function is well-written.

KyleFromKitware (changes requested):
Looks good for the most part, I suggested a few small changes.

test/core/DependencyUnitTests/CMakeLists.txt
```diff
@@ -123,11 +123,11 @@ function(read_expected_output_file_into_regex_array  EXPECTED_OUTPUT_FILE
...
     if (NOT "${EXPECTED_RAW_STR}" STREQUAL "")
       set(EXPECTED_REGEX_STR "${EXPECTED_RAW_STR}")
-      string(REPLACE "(" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}") 
-      string(REPLACE ")" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}") 
-      string(REPLACE "[" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}") 
-      string(REPLACE "]" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}") 
-      string(REPLACE "*" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}") 
+      string(REPLACE "[" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}")
+      string(REPLACE "]" "." EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}")
+      string(REPLACE "(" "[(]" EXPECTED_REGEX_STR "${EXPECTED_REGEX_STR}")
```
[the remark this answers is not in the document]
bartlettroscoe: I have not had good luck using `\\` to try to escape chars.  That is because it seems like one or more `\` will get removed as you pass arguments through layers of functions.  This behavior confuses me to the point that it is just not worth the tiny bit of extra checking in the output for differentiate a `[` from some arbitrary char compared to fighting with how many `\` are needed for a given situation.

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
```diff
@@ -472,30 +473,36 @@ endmacro()
...
+# variables for ``<packageName>`` described in `Variables defining the package
+# dependencies graph`_.  Note that the downstream/forward dependencies of
+# upstream packages for this package ``<packageName>`` are built up
+# incrementally.  (The forward dependency list vars are initialized to empty
+# in `tribits_prep_to_read_dependencies()`_.)
 #
-# See `Function call tree for constructing package dependency graph`_
+# See `Function call tree for constructing package dependency graph`_.
+#
+# **__Legacy varibles #63__**
```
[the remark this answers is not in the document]
bartlettroscoe: All of these sections will get removed when I complete #63 and remove all of the code and comments that mention `#63` like t his.

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
```diff
@@ -610,37 +599,46 @@ endfunction()
...
+function(tribits_append_forward_dep_packages  packageName  inputListType)
 
-  set(DEP_PKG_LIST_NAME "${PACKAGE_NAME}_${LIST_TYPE}")
+  set(depPkgListName "${packageName}_${inputListType}")
 
   assert_defined(${PROJECT_NAME}_ASSERT_MISSING_PACKAGES)
-  foreach(DEP_PKG ${${DEP_PKG_LIST_NAME}})
-    set(FWD_DEP_PKG_LIST_NAME "${DEP_PKG}_FORWARD_${LIST_TYPE}")
-    if (NOT DEFINED ${FWD_DEP_PKG_LIST_NAME})
+  foreach(depPkg ${${depPkgListName}})
```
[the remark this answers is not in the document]
bartlettroscoe: This was old code from 12+ years ago before CMake supported `IN LISTS`.  But note that I don't think `IN LISTS` works if the argument is a macro parameter, correct?  Therefore, it seems to be more robust to use the form without `IN LISTS` in some contexts.

test/core/DependencyUnitTests/CMakeLists.txt
bartlettroscoe: For consistency, sure.  But as we know, no behavior is altered at all for types like `BOOL` (unlike `PATH` and `FILEPATH`).

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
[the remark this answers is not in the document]
KyleFromKitware: The following works as expected:

```cmake
macro(print_list var)
  foreach(i IN LISTS ${var})
    message(STATUS "${i}")
  endforeach()
endmacro()

set(list a b c)
print_list(list)
```

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
bartlettroscoe: I know the issues.  The problem I have is that it is very difficult to work with global variables in any type of logical way.  When you have a set of  CMake code that is trying to coordinate to build a data-structure of top-level project regular variables (not INTERNAL cache vars), you have to jump through all types of hoops.  And if you have a stack of function calls and you want to be setting and updating a top-level project scope var, you have to propagate those manually using `set(... PARENT_SCOPE)`.  It is a terrible situation.

I guess a way to make a stack of CMake macro calls more safe would be to namespace all of the local variables with a unique prefix (e.g. based on the macro name like `tspd_` for `tribits_set_dep_packages()`) and that would avoid classes with local variables above and below the call stack.  The one exception is macro arguments where are not CMake variables (which is super confusing because `${<macroArg}` works just like for a real CMake variable).

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
bartlettroscoe: This is the deal with macro arguments where I changed from a function() to a macro().  It seems like the only way to use variables or macro arguments in `if()` statements that is robust to all refactorings is to just give up and use `"${<varOrMacroArgName>}"` instead of `<varOrMacroArgName>`. That was the case here (the code is broken if you replace this with `if (requiredOrOptional STREQUAL "REQUIRED")`).  Once I started using `if ( ...  "${<varOrMacroArgName>}"  ...)`, the code behaved much more robust to refactorings.

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
bartlettroscoe: I made the needed adjustments.  See discussion [above](https://github.com/TriBITSPub/TriBITS/pull/530#discussion_r996158789).

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
```diff
@@ -526,76 +533,58 @@ endmacro()
...
-  set(PACKAGE_ENABLE_VAR  ${PROJECT_NAME}_ENABLE_${PACKAGE_NAME})
+  set(legacyPackageDepsList "") # Legacy var #63
 
-  foreach(DEP_PKG ${${LIST_TYPE}})
-    if (TRIBITS_SET_DEP_PACKAGES_DEBUG_DUMP)
-      print_var(DEP_PKG)
-    endif()
-    if (${DEP_PKG} STREQUAL ${PACKAGE_NAME})
-      tribits_abort_on_self_dep("${PACKAGE_NAME}" "${LIST_TYPE}")
+  foreach(depPkg ${${inputListType}})
```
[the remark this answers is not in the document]
bartlettroscoe: This does not seem to exist in the final version of the code.  See:

https://github.com/TriBITSPub/TriBITS/blob/b64983e0449d768c03a5a05b1c3591016b54170a/tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake#L555

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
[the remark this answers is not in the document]
bartlettroscoe: But I will fix all of the cases that I can find in that file where IN LISTS can be used (because one would assume it might be more efficient)?

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
bartlettroscoe: Is it not true all of the things we are saying about macros are also true about local variables created in `Find<Package>.cmake` files and `<Package>Config.cmake` files?  When those files get processed, are then not run in the same local variable scope as the code calling `find_package()`?   I think suggests that all of this code needs to carefully namespace local variables.

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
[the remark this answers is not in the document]
bartlettroscoe: > The following works as expected:
> 
> ```cmake
> macro(print_list var)
>   foreach(i IN LISTS ${var})
>     message(STATUS "${i}")
>   endforeach()
> endmacro()
> 
> set(list a b c)
> print_list(list)
> ```

I guess that is true.  If `var` is the name of a list var, then `${var}` will evaluate to the name of the list var and things will work correctly.  I guess I was thinking about the behavior of macro args in `if()` statements not behaving the same was as regular vars.

test/core/DependencyUnitTests/CMakeLists.txt
bartlettroscoe: Resolved in new commit 515df6c7.

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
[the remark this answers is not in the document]
bartlettroscoe: All such cases in the file are resolved in new commit 0c611554.

bartlettroscoe:
@KyleFromKitware, can you please have a look at the remaining unresolved conversations above [here](https://github.com/TriBITSPub/TriBITS/pull/530#discussion_r996030071) and [here](https://github.com/TriBITSPub/TriBITS/pull/530#discussion_r996030760) and the approve the PR if you are satisfied?

KyleFromKitware (approved):

tribits/core/package_arch/TribitsReadDepsFilesCreateDepsGraph.cmake
bartlettroscoe: >  though I don't see it as much effort to just add `set(... PARENT_SCOPE)` at the end.

When you have as stack of subroutines that are building a  global data-structure all of this is just so messy.  It is not hard to come up with use cases where variables don't behave the way you think they should.  I think I would just rather namespace a small number of local vars within a macro in cases like this than to fight the CMake variable scoping and hiding system (see section "5.4. Potentially Surprising Behavior Of Variables" in the book "Professional CMake").  Otherwise, I try to use functions as much as possible to avoid all of these problems (and all routines that don't modify state should be functions).

Issue #534, opened 2022-10-21
Using a custom MPI library


ibaned:
I've found a solution for this issue that I will outline below, but I thought it was appropriate to describe here as its own GitHub issue. I use my own implementation of MPI for one rank with TriBITS-based projects: https://github.com/sandialabs/fbmpi.

When using TriBITS' MPI TPL, I couldn't specify my own `TPL_MPI_INCLUDE_DIRS` unless the call to `tribits_tpl_find_include_dirs_and_libraries` had a `REQUIRED_HEADERS` argument. We can provide this with `mpi.h`, because that is what all MPI codes include. Once that modification is made, I was able to specify the following at configuration time:
```
-DTPL_ENABLE_MPI=ON
-DTPL_MPI_INCLUDE_DIRS=$MY_INCLUDE_PATH
-DTPL_MPI_LIBRARIES=$MY_LIBRARY
-DMPI_LIBRARY_NAMES=fbmpi
```

This issue can form a useful anchor for related branches and pull requests with this change

ibaned:
@bartlettroscoe I might need your help with this too... as you point out it breaks normal Trilinos builds:

https://github.com/trilinos/Trilinos/pull/11175#discussion_r1001979683

Is there a way that the TriBITS MPI TPL can be set up to work both with the usual MPI compiler wrappers and with a specified include dir and library? Right now I'm running into this line which erases my `TPL_MPI_INCLUDE_DIRS` setting:

https://github.com/TriBITSPub/TriBITS/blob/2a389b7bf4203941a548a31d4e3cea0dbbfc2c3e/tribits/core/package_arch/TribitsTplFindIncludeDirsAndLibraries.cmake#L680

ibaned:
actually... maybe that file is different in my copy of SEACAS and latest TriBITS...

ibaned:
Nope, even with the latest snapshot of TriBITS this doesn't work

bartlettroscoe:
> https://github.com/TriBITSPub/TriBITS/blob/2a389b7bf4203941a548a31d4e3cea0dbbfc2c3e/tribits/core/package_arch/TribitsTplFindIncludeDirsAndLibraries.cmake#L680

The simple fix would be to change that line to:

```
      # Library has no header files, no user override, so just set them to null by default
      if ("${TPL_${TPL_NAME}_INCLUDE_DIRS}" STREQUAL "")
[...]

Issue #533, opened 2022-10-19
Find NetCDF and HDF5 using only CMake config files


ibaned:
Since recent versions of NetCDF and HDF5 both install good CMake config files and export proper CMake targets, it would be great (actually, it is necessary for me) if TriBITS could bring in these TPLs using only these CMake config files and targets. In particular, TriBITS should do the equivalent of the following:
```cmake
find_package(hdf5 CONFIG REQUIRED)
find_package(netCDF CONFIG REQUIRED)
target_link_libraries(downstream PUBLIC netCDF::netcdf)
```
and only require `hdf5_ROOT` and `netCDF_ROOT` to be defined during CMake configuration.
I'm currently doing a lot of work to avoid having SEACAS and Trilinos in my application build due to not being able to do this.

bartlettroscoe:
That is that idea.  See [Creating FindTPL<tplName>.cmake using find_package() with IMPORTED targets](https://tribitspub.github.io/TriBITS/users_guide/index.html#creating-findtpl-tplname-cmake-using-find-package-with-imported-targets).

What are the minimum versions of HDF5 and NetCDF needed to supply these package config files?  I think you would have to feed those versions into the `find_package(...)` calls.  And updated versions of HDF5 and NetCDF would have to be installed **everywhere** to make a change like this.  (But it is possible to upgrade the TriBITS `FindTPL<tplName>.cmake` modules to work with both old and new versions of these TPLs to maintain backward compatibility.)

We will need to coordinate with @gsjaardema and this will need to get rolled out and we should likely complete the [transition to CMake 3.22 as the minimum version](https://github.com/trilinos/Trilinos/issues/10355) to be safe.

ibaned:
It looks like at least HDF5 1.10.8 and netCDF 4.3.0 export a CMake config file at least... I'm not sure exactly what the compatibility range is.

However, I think we could do the following right now: change `FindTPL<tplName>.cmake` to attempt first to use CMake config files and only if that fails fall back to other methods. Does that sound reasonable? If it would be a change you would accept I could work on a PR.

ibaned:
I tried the following changes:

https://github.com/sandialabs/seacas/pull/336/files

And just using `tribits_extpkg_create_imported_all_libs_target_and_config_file` and returning doesn't seem to work as I would expect:
```
Processing enabled external package/TPL: Netcdf (enabled by SEACASExodus, disable with -DTPL_ENABLE_Netcdf=OFF)
[...]

PR #535, opened 2022-10-26
Update MPI, HDF5, and Netcdf TriBITS find modules (#533, #534)

bartlettroscoe:
## Description

These commits address #533 and #534

These commits were pulled off of the branch from:

* https://github.com/trilinos/Trilinos/pull/11175

(which also matches the patches in https://github.com/sandialabs/seacas/pull/336).

This seems to maintain near perfect backwards compatibility.
 
## Tasks

* [x] Test and merge https://github.com/trilinos/Trilinos/pull/11175
* [x] Test and merge https://github.com/sandialabs/seacas/pull/336
* [x] Get @KyleFromKitware to review
* [x] Address review comments with new commits
* [x] ~~Add test for change to `tribits/core/package_arch/TribitsTplFindIncludeDirsAndLibraries.cmake`~~ ... Cost of adding a test for this is not worth the value (see [below](https://github.com/TriBITSPub/TriBITS/pull/535#issuecomment-1295070533)).

### DIFF
--- a/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/tribits/common_tpls/FindTPLHDF5.cmake
@@ -4,17 +4,18 @@
 
 if (Netcdf_ALLOW_MODERN)
 
-  #set(minimum_modern_HDF5_version 4.7.4)
+  set(minimum_modern_HDF5_version 1.13.2)
   print_var(Netcdf_ALLOW_MODERN)
   message("-- Using find_package(HDF5 ${minimum_modern_HDF5_version} CONFIG) ...")
   find_package(HDF5  ${minimum_modern_HDF5_version}  CONFIG)
   if (HDF5_FOUND)
     message("-- Found HDF5_CONFIG=${HDF5_CONFIG}")
     message("-- Generating Netcdf::all_libs and NetcdfConfig.cmake")
+    message("-- HDF5_EXPORT_LIBRARIES=${HDF5_EXPORT_LIBRARIES}")
     tribits_extpkg_create_imported_all_libs_target_and_config_file(
       HDF5
       INNER_FIND_PACKAGE_NAME  HDF5
-      IMPORTED_TARGETS_FOR_ALL_LIBS   HDF5::HDF5)
+      IMPORTED_TARGETS_FOR_ALL_LIBS   ${HDF5_EXPORT_LIBRARIES})
   endif()
 
 endif()
--- a/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/tribits/common_tpls/FindTPLHDF5.cmake
@@ -2,78 +2,97 @@
 # See associated tribits/Copyright.txt file for copyright and license! #
 ########################################################################
 
-#
-# First, set up the variables for the (backward-compatible) TriBITS way of
-# finding HDF5.  These are used in case find_package(HDF5 ...) is not called
-# or does not find HDF5.  Also, these variables need to be non-null in order
-# to trigger the right behavior in the function
-# tribits_tpl_find_include_dirs_and_libraries().
-#
+if (Netcdf_ALLOW_MODERN)
 
-set(REQUIRED_HEADERS hdf5.h)
-set(REQUIRED_LIBS_NAMES hdf5)
+  #set(minimum_modern_HDF5_version 4.7.4)
+  print_var(Netcdf_ALLOW_MODERN)
+  message("-- Using find_package(HDF5 ${minimum_modern_HDF5_version} CONFIG) ...")
+  find_package(HDF5  ${minimum_modern_HDF5_version}  CONFIG)
+  if (HDF5_FOUND)
+    message("-- Found HDF5_CONFIG=${HDF5_CONFIG}")
+    message("-- Generating Netcdf::all_libs and NetcdfConfig.cmake")
+    tribits_extpkg_create_imported_all_libs_target_and_config_file(
+      HDF5
+      INNER_FIND_PACKAGE_NAME  HDF5
+      IMPORTED_TARGETS_FOR_ALL_LIBS   HDF5::HDF5)
+  endif()
 
-if (HDF5_REQUIRE_FORTRAN)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_fortran)
 endif()
 
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
-endif()
+if (NOT TARGET HDF5::all_libs)
 
-if (TPL_ENABLE_Netcdf)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_hl)
-endif()
+  # First, set up the variables for the (backward-compatible) TriBITS way of
+  # finding HDF5.  These are used in case find_package(HDF5 ...) is not called
+  # or does not find HDF5.  Also, these variables need to be non-null in order
+  # to trigger the right behavior in the function
+  # tribits_tpl_find_include_dirs_and_libraries().
 
-#
-# Second, search for HDF5 components (if allowed) using the standard
-# find_package(HDF5 ...).
-#
-tribits_tpl_allow_pre_find_package(HDF5  HDF5_ALLOW_PREFIND)
-if (HDF5_ALLOW_PREFIND)
+  set(REQUIRED_HEADERS hdf5.h)
+  set(REQUIRED_LIBS_NAMES hdf5)
 
-  message("-- Using find_package(HDF5 ...) ...")
-
-  set(HDF5_COMPONENTS C)
   if (HDF5_REQUIRE_FORTRAN)
-    list(APPEND HDF5_COMPONENTS Fortran)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_fortran)
   endif()
 
   if (TPL_ENABLE_MPI)
-    set(HDF5_PREFER_PARALLEL TRUE)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
   endif()
 
-  find_package(HDF5 COMPONENTS ${HDF5_COMPONENTS})
-
-  # Make sure that HDF5 is parallel.
-  if (TPL_ENABLE_MPI AND NOT HDF5_IS_PARALLEL)
-    message(FATAL_ERROR "Trilinos is configured for MPI, HDF5 is not.
-    Did CMake find the correct libraries?
-    Try setting HDF5_INCLUDE_DIRS and/or HDF5_LIBRARY_DIRS explicitly.
-    ")
+  if (TPL_ENABLE_Netcdf)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_hl)
   endif()
 
-  if (HDF5_FOUND)
-    # Tell TriBITS that we found HDF5 and there no need to look any further!
-    set(TPL_HDF5_INCLUDE_DIRS ${HDF5_INCLUDE_DIRS} CACHE PATH
-      "HDF5 include dirs")
-    set(TPL_HDF5_LIBRARIES ${HDF5_LIBRARIES} CACHE FILEPATH
-      "HDF5 libraries")
-    set(TPL_HDF5_LIBRARY_DIRS ${HDF5_LIBRARY_DIRS} CACHE PATH
-      "HDF5 library dirs")
+  #
+  # Second, search for HDF5 components (if allowed) using the standard
+  # find_package(HDF5 ...).
+  #
+  tribits_tpl_allow_pre_find_package(HDF5  HDF5_ALLOW_PREFIND)
+  if (HDF5_ALLOW_PREFIND)
+
+    message("-- Using find_package(HDF5 ...) ...")
+
+    set(HDF5_COMPONENTS C)
+    if (HDF5_REQUIRE_FORTRAN)
+      list(APPEND HDF5_COMPONENTS Fortran)
+    endif()
+
+    if (TPL_ENABLE_MPI)
+      set(HDF5_PREFER_PARALLEL TRUE)
+    endif()
+
+    find_package(HDF5 COMPONENTS ${HDF5_COMPONENTS})
+
+    # Make sure that HDF5 is parallel.
+    if (TPL_ENABLE_MPI AND NOT HDF5_IS_PARALLEL)
+      message(FATAL_ERROR "Trilinos is configured for MPI, HDF5 is not.
+      Did CMake find the correct libraries?
+      Try setting HDF5_INCLUDE_DIRS and/or HDF5_LIBRARY_DIRS explicitly.
+      ")
+    endif()
+
+    if (HDF5_FOUND)
+      # Tell TriBITS that we found HDF5 and there no need to look any further!
+      set(TPL_HDF5_INCLUDE_DIRS ${HDF5_INCLUDE_DIRS} CACHE PATH
+        "HDF5 include dirs")
+      set(TPL_HDF5_LIBRARIES ${HDF5_LIBRARIES} CACHE FILEPATH
+        "HDF5 libraries")
+      set(TPL_HDF5_LIBRARY_DIRS ${HDF5_LIBRARY_DIRS} CACHE PATH
+        "HDF5 library dirs")
+    endif()
+
   endif()
+
+  #
+  # Third, call tribits_tpl_find_include_dirs_and_libraries()
+  #
+  tribits_tpl_find_include_dirs_and_libraries( HDF5
+    REQUIRED_HEADERS ${REQUIRED_HEADERS}
+    REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES}
+    )
+  # NOTE: If find_package(HDF5 ...) was called and successfully found HDF5, then
+  # tribits_tpl_find_include_dirs_and_libraries() will use the already-set
+  # variables TPL_HDF5_INCLUDE_DIRS and TPL_HDF5_LIBRARIES and then print them
+  # out (and set some other standard variables as well).  This is the final
+  # "hook" into the TriBITS TPL system.
 
 endif()
-
-#
-# Third, call tribits_tpl_find_include_dirs_and_libraries()
-#
-tribits_tpl_find_include_dirs_and_libraries( HDF5
-  REQUIRED_HEADERS ${REQUIRED_HEADERS}
-  REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES}
-  )
-# NOTE: If find_package(HDF5 ...) was called and successfully found HDF5, then
-# tribits_tpl_find_include_dirs_and_libraries() will use the already-set
-# variables TPL_HDF5_INCLUDE_DIRS and TPL_HDF5_LIBRARIES and then print them
-# out (and set some other standard variables as well).  This is the final
-# "hook" into the TriBITS TPL system.
--- a/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/tribits/common_tpls/FindTPLHDF5.cmake
@@ -2,7 +2,7 @@
 # See associated tribits/Copyright.txt file for copyright and license! #
 ########################################################################
 
-set(INTERNAL_IS_MODERN FALSE)
+set(HDF5_INTERNAL_IS_MODERN FALSE)
 
 if (Netcdf_ALLOW_MODERN)
 
@@ -18,12 +18,12 @@ if (Netcdf_ALLOW_MODERN)
       HDF5
       INNER_FIND_PACKAGE_NAME  HDF5
       IMPORTED_TARGETS_FOR_ALL_LIBS   ${HDF5_EXPORT_LIBRARIES})
-    set(INTERNAL_IS_MODERN TRUE)
+    set(HDF5_INTERNAL_IS_MODERN TRUE)
   endif()
 
 endif()
 
-set(HDF5_FOUND_MODERN_CONFIG_FILE ${INTERNAL_IS_MODERN} CACHE BOOL "True if HDF5 was found by the modern method" FORCE)
+set(HDF5_FOUND_MODERN_CONFIG_FILE ${HDF5_INTERNAL_IS_MODERN} CACHE INTERNAL "True if HDF5 was found by the modern method")
 
 if (NOT TARGET HDF5::all_libs)
 
--- a/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/tribits/common_tpls/FindTPLHDF5.cmake
@@ -2,6 +2,8 @@
 # See associated tribits/Copyright.txt file for copyright and license! #
 ########################################################################
 
+set(INTERNAL_IS_MODERN FALSE)
+
 if (Netcdf_ALLOW_MODERN)
 
   set(minimum_modern_HDF5_version 1.13.2)
@@ -16,9 +18,12 @@ if (Netcdf_ALLOW_MODERN)
       HDF5
       INNER_FIND_PACKAGE_NAME  HDF5
       IMPORTED_TARGETS_FOR_ALL_LIBS   ${HDF5_EXPORT_LIBRARIES})
+    set(INTERNAL_IS_MODERN TRUE)
   endif()
 
 endif()
+
+set(HDF5_FOUND_MODERN_CONFIG_FILE ${INTERNAL_IS_MODERN} CACHE BOOL "True if HDF5 was found by the modern method" FORCE)
 
 if (NOT TARGET HDF5::all_libs)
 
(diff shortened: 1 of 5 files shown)


tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -2,78 +2,103 @@
...
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
-endif()
+if (Netcdf_ALLOW_MODERN)
+
+  set(minimum_modern_HDF5_version 1.13.2)
+  print_var(Netcdf_ALLOW_MODERN)
+  message("-- Using find_package(HDF5 ${minimum_modern_HDF5_version} CONFIG) ...")
+  find_package(HDF5  ${minimum_modern_HDF5_version}  CONFIG)
+  if (HDF5_FOUND)
```
bartlettroscoe, line 13: @ibaned and @gsjaardema, following on from the comments https://github.com/sandialabs/seacas/pull/336#pullrequestreview-1157178993, https://github.com/trilinos/Trilinos/pull/11175#discussion_r1006168735, and https://github.com/trilinos/Trilinos/pull/11175#discussion_r1006169418, is it worth considering if we should try to change the `find_package([HDF5|netCDF] <minVersion> CONFIG)` calls which require a specific version to `find_package([HDF5|netCDF] <minVersion>... CONFIG)` to only require a minimum version?

Wait, the book "Professional CMake" says:

> The optional version argument indicates that the package must be of the specified version or higher, or match exactly if `EXACT` is also given.

so it would seem that a single version number **is** supposed to already be interpreted as the minimum version, not the exact version.  (That is not at all clear from my reading of the [official find_package() documentation](https://cmake.org/cmake/help/v3.25/command/find_package.html).

bartlettroscoe:
Okay, given [above](https://github.com/TriBITSPub/TriBITS/pull/535#pullrequestreview-1157314120), hopefully these new `FindTPL[HDF5|Netcdf].cmake` modules will work with newer releases of HDF5 and NetCDF as well.

I will merge this after the merge of:

* https://github.com/trilinos/Trilinos/pull/11175

We can always tweak this later if needed.

bartlettroscoe:
With the merge of:

* https://github.com/trilinos/Trilinos/pull/11175
* https://github.com/sandialabs/seacas/pull/336

we can finally merge this PR.

tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -44,98 +44,158 @@
 # in order to trigger the right behavior in the function
 # tribits_tpl_find_include_dirs_and_libraries().
 #
-
-set(REQUIRED_HEADERS netcdf.h)
-set(REQUIRED_LIBS_NAMES netcdf)
-
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} pnetcdf)
+if (${CMAKE_VERSION} GREATER "3.13")
```
KyleFromKitware, line 40:
```cmake
if(NOT CMAKE_VERSION VERSION_LESS "3.13")
```

Though, you're now requiring CMake 3.17 or so right? So is this even needed?

tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -44,98 +44,158 @@
...
+    # do find_dependency(hdf5)
+    add_library(Netcdf::all_libs  INTERFACE  IMPORTED  GLOBAL)
+    target_link_libraries(Netcdf::all_libs  INTERFACE  netCDF::netcdf)
+    set(configFileStr "")
+    string(APPEND configFileStr
+      "include(CMakeFindDependencyMacro)\n" )
+    string(APPEND configFileStr
+      "set(netCDF_DIR \"${netCDF_DIR}\")\n" )
+    string(APPEND configFileStr
+      "find_dependency(netCDF ${minimum_modern_netCDF_version} CONFIG)\n"
```
KyleFromKitware: These three `string(APPEND)` calls could be collapsed into one.

tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -2,78 +2,97 @@
...
-  find_package(HDF5 COMPONENTS ${HDF5_COMPONENTS})
-
-  # Make sure that HDF5 is parallel.
-  if (TPL_ENABLE_MPI AND NOT HDF5_IS_PARALLEL)
-    message(FATAL_ERROR "Trilinos is configured for MPI, HDF5 is not.
-    Did CMake find the correct libraries?
-    Try setting HDF5_INCLUDE_DIRS and/or HDF5_LIBRARY_DIRS explicitly.
-    ")
+  if (TPL_ENABLE_Netcdf)
+    set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_hl)
```
KyleFromKitware, line 48: You could use `list(APPEND)` here.

tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -44,98 +44,158 @@
...
+  #
+  # Second, search for Netcdf components (if allowed) using the standard
+  # find_package(NetCDF ...).
+  #
+  tribits_tpl_allow_pre_find_package(Netcdf  Netcdf_ALLOW_PREFIND)
+  if (Netcdf_ALLOW_PREFIND)
+
+    message("-- Using find_package(Netcdf ...) ...")
+
+    set(CMAKE_MODULE_PATH
```
KyleFromKitware, line 88: You could use `list(APPEND)` here.

tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -16,10 +18,13 @@ if (Netcdf_ALLOW_MODERN)
       HDF5
       INNER_FIND_PACKAGE_NAME  HDF5
       IMPORTED_TARGETS_FOR_ALL_LIBS   ${HDF5_EXPORT_LIBRARIES})
+    set(INTERNAL_IS_MODERN TRUE)
   endif()
 
 endif()
 
+set(HDF5_FOUND_MODERN_CONFIG_FILE ${INTERNAL_IS_MODERN} CACHE BOOL "True if HDF5 was found by the modern method" FORCE)
```
KyleFromKitware: `${INTERNAL_IS_MODERN}` should be quoted to ensure that `set()` receives a value... though I see from above that it will always have a value, so perhaps it doesn't matter. Still, quoting it makes this explicit.

KyleFromKitware (changes requested):

tribits/common_tpls/FindTPLNetcdf.cmake
bartlettroscoe: That is a good point.  This file only gets pulled in for TriBITS projects where CMake 3.17+ is required (and soon will be 3.22+).

tribits/common_tpls/FindTPLNetcdf.cmake
bartlettroscoe: This code is not in the final version of the PR.  I left in this this intermediate commit on purpose to document how TriBITS TPL dependencies solved this problem with very minimal code.

tribits/common_tpls/FindTPLHDF5.cmake
bartlettroscoe: Good point. This was old code so we were just trying to avoid breaking anything.  Also, hopefully, this specialized code will disappear at some point.

tribits/common_tpls/FindTPLHDF5.cmake
bartlettroscoe: This is a good point I never considered. If `INTERNAL_IS_MODERN` is empty (or undefined), this statement just produces a regular list variable with the value:

```
"CACHE;BOOL;True if HDF5 was found by the modern method;FORCE"
```

Is that correct?

bartlettroscoe: > So it turns out that if it's empty or undefined, it will simply set a cache variable with an empty value. However, readers won't know this just from glancing at the code, and for that reason I still recommend quoting it.

Then what is the benefit of quoting the `${val}` argument in `set(<var> ${val} CACHE ...)` if it does not impact behavior?  In my example [above](https://github.com/TriBITSPub/TriBITS/pull/535#discussion_r1007042793), quoting the argument can  change behavior.

KyleFromKitware: > then what is the benefit of quoting the ${val} argument in set(<var> ${val} CACHE ...) if it does not impact behavior?

It assures the reader that there won't be a problem. I had to test the behavior of `set(... CACHE)` to check that it would work without quoting, whereas I knew instantly just by looking at it that it would work *with* quoting. I believe it's better to be explicit than implicit. Just a thought.

tribits/common_tpls/FindTPLNetcdf.cmake
bartlettroscoe: Because this file is so hard to fully test in all of the use cases, I am not going to change this code at this point.  (In the future, it should all be stripped out.)

tribits/common_tpls/FindTPLHDF5.cmake
bartlettroscoe: Because this file is so hard to fully test in all of the use cases, I am not going to change this code at this point. (In the future, it should all be stripped out.)

bartlettroscoe:
I am looking over how to test [`7fd57e5` (#535)](https://github.com/TriBITSPub/TriBITS/pull/535/commits/7fd57e5d254a8f2769a4b6732776178515483f72) and it will require quite a lot of work and a lot of code.  That code in that file has not changed for years and is unlikely to ever change. (Even the large refactoring in #299 did not touch that code at all.  Therefore, I have determined that the cost of adding a test for that change is larger than the value it will have.  (This is not a decision I take likely.)

tribits/common_tpls/FindTPLNetcdf.cmake
bartlettroscoe: This is resolved in the new commit [`4127bf3` (#535)](https://github.com/TriBITSPub/TriBITS/pull/535/commits/4127bf3e4cda37614da92489ce7dba643dfabf45).

rabartlett1972 (approved):
As the other changes are very hard to test as noted, I will approve this PR to be merged.

In the future, we need to be reviewing these changes in the Trilinos and SEACAS PRs where they can be tested.

tribits/common_tpls/FindTPLHDF5.cmake
bartlettroscoe: That is a good point.  There really needs to be a CMake linter to catch issues like this automatically.  We should not expend too much time in PR reviews dealing with low-level language issues like this.  That is why they make linters in the first place.

In the future, we will need your help reviewing these types of changes when they are being created and tested in the Trilinos github repo.  Once we being this back to TriBITS (as the official repo for these common TPLs) we can't test them easily.

tribits/common_tpls/FindTPLNetcdf.cmake
```diff
@@ -37,105 +37,128 @@
...
+cmake_policy(SET CMP0074 NEW)
 
-set(REQUIRED_HEADERS netcdf.h)
-set(REQUIRED_LIBS_NAMES netcdf)
+set(Netcdf_ALLOW_MODERN FALSE CACHE BOOL "Allow finding Netcdf as a modern CMake config file with exported targets (and only this way)")
 
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} pnetcdf)
-endif()
+if (Netcdf_ALLOW_MODERN AND HDF5_FOUND_MODERN_CONFIG_FILE)
```
gsjaardema, line 44: After seeing how to use this in SEACAS -- 
This does not seem to work *if* HDF5 is added as a dependency only because of netCDF since at this point, I don't think that we have yet found HDF5, so `HDF5_FOUND_MODERN_CONFIG_FILE` will not be set?

At least for SEACAS, I think that the `Netcdf_ALLOW_MODERN` should drive this and determine whether to find HDF5 the modern/legacy way?  Does that make sense?

Or, do I reorder things such that CGNS (which depends on HDF5) comes first and then netCDF?
@bartlettroscoe @ibaned

bartlettroscoe: > Or, do I reorder things such that CGNS (which depends on HDF5) comes first and then netCDF?

@gsjaardema, you specify the dependencies as: HDF5 <-- CGNS <-- Netcdf

See:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#tpldefsdir-findtpl-tplname-dependencies-cmake

It looks like GGNS already has a required dependency on HDF5 as per:

https://github.com/TriBITSPub/TriBITS/blob/9fd56d2afdf1c8e38da30afda1fbb7b7c393e28c/tribits/common_tpls/FindTPLCGNSDependencies.cmake#L1-L2

But I had no idea that Netcdf depended on CGNS?

Also not that, currently, TriBITS does **not** automatically enable upstream TPL dependencies for an enabled TPL.  So the configure script will need to manually enable all of the TPLs it wants.  (See NOTES at bottom of slide 16 "TriBITS External Packages/TPLs Dependencies (New!)" in the presentation [TriBITS Modernization Update](https://trilinos.github.io/pages/community/trilinos_user_meetings/TUG_2022/2022-10-27/01_TribitsModernization_TUG_2022-10-27.pdf) I gave last Thursday, Oct 27, 2022.)

gsjaardema:
NetCDF doesn't depend on CGNS.  They both depend on HDF5, so I thought that maybe configuring CGNS first would then set the `MODERN` cmake for HDF5 which would then allow the `if (Netcdf_ALLOW_MODERN AND HDF5_FOUND_MODERN_CONFIG_FILE)` to work...

If netCDF is configured before CGNS, then that if won't be triggered because it is netCDF that triggers HDF5 if I am understanding things correctly (probably not)

bartlettroscoe:
> If netCDF is configured before CGNS, then that if won't be triggered because it is netCDF that triggers HDF5 if I am understanding things correctly (probably not)

@gsjaardema, the HDF5 TPL will always be configured before Netcdf TPL, because the TPLs are processed in the order they are listed in the [TPLsList.cmake](https://github.com/sandialabs/seacas/blob/master/TPLsList.cmake) file which is the case with SEACAS as shown at:

https://github.com/sandialabs/seacas/blob/a9634da294d7d4bc64145fa33969de6be10d31b6/TPLsList.cmake#L60-L62

gsjaardema:
This is a portion of the output of my CMake configuration:
```
Processing enabled external package/TPL: Netcdf (enabled explicitly, disable with -DTPL_ENABLE_Netcdf=OFF)
-- Using find_package(Netcdf ...) ...
--      NetCDF_ROOT is /Users/gdsjaar/src/seacas
--      netCDF_CONFIG is /Users/gdsjaar/src/seacas/lib/cmake/netCDF/netCDFConfig.cmake
-- NetCDF requires HDF5
-- NetCDF depends on HDF5
-- Updating NetCDF_LIBRARIES and NetCDF_INCLUDE_DIRS
-- Found HDF5 CMake configuration file ( directory /Users/gdsjaar/src/seacas/cmake/hdf5-config.cmake )
-- HDF5 Version: 1.10.9
--      HDF5_INCLUDE_DIRS      =/Users/gdsjaar/src/seacas/include
--      HDF5_LIBRARY_TARGETS   =hdf5_hl;hdf5
--      HDF5_LIBRARIES         =/Users/gdsjaar/src/seacas/lib/libhdf5_hl.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5.dylib
--      HDF5_LIBRARIES_EXPORT  =/Users/gdsjaar/src/seacas/lib/libhdf5_hl.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5.dylib
--      HDF5_LINK_LIBRARIES    =
--      HDF5_IS_PARALLEL       =NO
-- Found the following HDF5 component libraries
--      hdf5
--      hdf5_hl
--      HDF5 Components not found: static;shared;CXX;Fortran;CXX_HL;Fortran_HL;Java;Tools
--      HDF5_TOOLS_FOUND: h52gif;h5copy;h5debug;h5diff;h5dump;h5import;h5jam;h5ls;h5mkgrp;h5stat
-- Found HDF5: /Users/gdsjaar/src/seacas/include
-- NetCDF does not require PNetCDF
-- Found NetCDF: /Users/gdsjaar/src/seacas/lib/libnetcdf.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5_hl.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5.dylib
-- NetCDF Version: 4.9.0
--      NetCDF_NEEDS_HDF5        = yes
--      NetCDF_NEEDS_PNetCDF     = no
--      NetCDF_PARALLEL          = False
--      NetCDF_INCLUDE_DIRS      = /Users/gdsjaar/src/seacas/include;/Users/gdsjaar/src/seacas/include
--      NetCDF_LIBRARIES         = /Users/gdsjaar/src/seacas/lib/libnetcdf.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5_hl.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5.dylib
--      NetCDF_BINARIES          = ncdump;ncgen;nccopy
-- Netcdf_LIBRARY_NAMES='netcdf'
-- TPL_Netcdf_LIBRARIES='/Users/gdsjaar/src/seacas/lib/libnetcdf.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5_hl.dylib;/Users/gdsjaar/src/seacas/lib/libhdf5.dylib'
-- TPL_Netcdf_INCLUDE_DIRS='/Users/gdsjaar/src/seacas/include;/Users/gdsjaar/src/seacas/include'
-- TPL_Netcdf_PARALLEL is False
```
Doesn't that show that Netcdf is being found and then since netcdf depends on HDF5, then HDF5 is found/processed?

Note that HDF5 is an optional dependency wherever it is mentioned in the SEACAS `Dependencies.cmake`:
```
packages/seacas/libraries/exodus/cmake/Dependencies.cmake:  LIB_OPTIONAL_TPLS Pthread HDF5 Pnetcdf MPI
```
Thanks for trying to explain.

bartlettroscoe:
@gsjaardema, can you please attach the full cmake STDOUT+STDERR?

What are the dependencies of Pnetcdf?  Do we need to add `FindTPL<tplName>Dependencies.cmake` entries for those as well?

tribits/common_tpls/FindTPLHDF5.cmake
```diff
@@ -2,78 +2,103 @@
...
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} hdf5_fortran)
-endif()
+set(HDF5_INTERNAL_IS_MODERN FALSE)
 
-if (TPL_ENABLE_MPI)
-  set(REQUIRED_LIBS_NAMES ${REQUIRED_LIBS_NAMES} z)
-endif()
+if (Netcdf_ALLOW_MODERN)
+
+  set(minimum_modern_HDF5_version 1.13.2)
```
gsjaardema, line 9: @ibaned @bartlettroscoe 
Sorry for not noticing this prior to merging.  
What is needed in the HDF5-1.10.X and HDF5-1.12.X versions to make them usable for this.  Both of those versions are still in development, so it would be good to fix their CMake configurations if possible.

I can either contact the HDF5 group with the suggested changes, or one of you can also.  Note that we have priority support with HDF5, so they will listen and most likely make useful changes that we suggest.

gsjaardema:
> What are the dependencies of Pnetcdf? Do we need to add `FindTPL<tplName>Dependencies.cmake` entries for those as well?

PnetCDF only depends on MPI and netCDF optionally depends on PnetCDF in a parallel build.  SEACAS only accesses / depends on PnetCDF if netCDF was configured in parallel and with PnetCDF capability enabled.  There isn't any direct access from SEACAS that bypassed netCDF and hits PnetCDF.

bartlettroscoe:
> PnetCDF only depends on MPI and netCDF optionally depends on PnetCDF in a parallel build. SEACAS only accesses / depends on PnetCDF if netCDF was configured in parallel and with PnetCDF capability enabled. There isn't any direct access from SEACAS that bypassed netCDF and hits PnetCDF.

Then this dependency between the TriBITS `Pnetcdf` and `Netcdf` TPLs should be expressed by updating the file https://github.com/sandialabs/seacas/blob/master/cmake/tribits/common_tpls/FindTPLNetcdfDependencies.cmake to be:

```
tribits_extpkg_define_dependencies( Netcdf
  DEPENDENCIES  Pnetcdf HDF5)
```

In case we do turn on TPL dependencies by default, this might get changed to:

```
set(Netcdf_Pnetcdf_dep "")
if (TPL_ENABLE_MPI)
  set(Netcdf_Pnetcdf_dep  Pnetcdf)
endif()
tribits_extpkg_define_dependencies( Netcdf
  DEPENDENCIES  ${Netcdf_Pnetcdf_dep} HDF5)
```

That would mean that if the user set `TPL_ENABLE_Netcdf=ON` and `TPL_ENABLE_MPI=ON` on input, then TriBITS would try to  automatically enable `Pnetcdf` (once we turn on indirect upstream TPL dependency enables).

How common is it to enable MPI but not build `Netcdf` with `Pnetcdf`?  That is, in the future, do we want the user setting `TPL_ENABLE_Netcdf=ON` and `TPL_ENABLE_MPI=ON` to automatically enable `Pnetcdf`?  Note that the user will always be able to set `TPL_ENABLE_Pnetcdf=OFF` and the `Pnetcdf` TPL will not get enabled.  Is that logical behavior?  What is the least surprising behavior?

gsjaardema:
> ... deleted...
> Then this dependency between the TriBITS `Pnetcdf` and `Netcdf` TPLs should be expressed by updating the file https://github.com/sandialabs/seacas/blob/master/cmake/tribits/common_tpls/FindTPLNetcdfDependencies.cmake to be:
> 
> ```
> tribits_extpkg_define_dependencies( Netcdf
>   DEPENDENCIES  Pnetcdf HDF5)
> ```

I still confused as to why the netCDF configuration isn't the place where this information is handled.  That is the current behavior and it seems to be working in most situations...  

netCDF has a few optional dependencies depending on how it is configured:
- hdf5
- pnetcdf
- dap, curl, nczarr, 
- compression: szip, bz2, zlib, zstd, blosc
- libxml2
- mpi

We will also have that issue with HDF5 soon if we start adding some of the advanced HDF5 capabilties (VOLs will add optional dependencies); multiple compression library options.

As I see it, only the netCDF configuration files can fully know what optional dependencies are present in the library...

bartlettroscoe:
> As I see it, only the netCDF configuration files can fully know what optional dependencies are present in the library...

The `netCDFConfig.cmake`	 file is broken.  It does not contain a call to `find_dependency(HDF5)`.  @ibaned knows more about that.

TriBITS allows us to fix issues with these external packages when they are not doing the correct behavior.

gsjaardema:
> The `netCDFConfig.cmake` file is broken. It does not contain a call to `find_dependency(HDF5)`. @ibaned knows more about that.
> 
> TriBITS allows us to fix issues with these external packages when they are not doing the correct behavior.

Ok, that was the missing piece for me --

ibaned:
Hi! I'm coming in late to the conversation, I gather essentially that this PR broke something? Could we maybe open an issue on the SEACAS or TriBITS repositories just to track what broke and I can get a better understanding?

gsjaardema:
@ibaned Nothing broke that I am aware of, I am just not sure it is working the way it should -- I am unable to get the `MODERN` config to work due to the way that the HDF5 dependency is currently added via netCDF and not as a direct SEACAS dependency.

ibaned:
Ah I see! If it helps, here is the CMake configuration that I'm using:
```
  "-DSeacas_ENABLE_Fortran=OFF"
  "-DSeacas_ENABLE_SEACASExodus=ON"
  "-DSEACASExodus_ENABLE_MPI=ON"
  "-DSEACASExodus_ENABLE_SHARED=OFF"
  "-DSeacas_ENABLE_SEACASAprepro_lib=ON"
  "-DTPL_ENABLE_HDF5=ON"
  "-DNetcdf_ALLOW_MODERN=ON"
  "-DnetCDF_ROOT=${CAPP_INSTALL_ROOT}/netcdf-c"
  "-DHDF5_ROOT=${CAPP_INSTALL_ROOT}/hdf5"
  ${SEACAS_MPI_OPTIONS}
  "-DSeacas_VERBOSE_CONFIGURE=OFF"
  "-DTRIBITS_TPL_FIND_INCLUDE_DIRS_AND_LIBRARIES_VERBOSE=OFF"
  "-DBUILD_SHARED_LIBS=OFF"
```

gsjaardema:
It is probably the `"-DTPL_ENABLE_HDF5=ON"` that does it.

Also, do you know what HDF5-1.10.X is missing to allow it to use the `MODERN` method...

ibaned:
Maybe nothing! I was being overly conservative in listing just the versions installed on my system.

bartlettroscoe:
@gsjaardema, can we set up a short MS Teams meeting so I can explain to you what is happening and how TriBITS is fixing this?

bartlettroscoe:
> > TriBITS allows us to fix issues with these external packages when they are not doing the correct behavior.
> 
> Ok, that was the missing piece for me --

@gsjaardema, the key is that when setting `TPL_ENABLE_HDF5=ON`, TriBITS generates a file `NetcdfConfig.cmake` in the build and install trees that basically does:

```
set(HDF5_DIR ...)
find_dependency(HDF5)
set(netCDF_DIR ...)
find_dependency(netCDF)
add_library(Netcdf::all_libs IMPORTED INTERFACE)
...
```

Look at the file:

```
<buildDir>/external_packages/Netcdf/NetcdfConfig.cmake
```

and you will see this.

And it is possible to extend this TriBITS support to automatically fix other situations as well.

Issue #429, opened 2021-11-03
Design and implement strategy for deprecating TriBITS code


bartlettroscoe:
**Parent Epic:** #411

## Description

As noted in https://github.com/trilinos/Trilinos/pull/9894#discussion_r742142095, we need to consider how to handle deprecated functions and macros in TriBITS.    I don't think it is desirable to just use raw `message(DEPRECATION <msg>)` since that might generate thousands of lines of output if the deprecated function/macro is used in a loop.  But it would be nice to have a system where the user can decide to show deprecated code or not.  Also, it would be good to allow the separation of deprecation warnings for TriBITS from deprecations for other CMake modules and packages and even the user's own project functions.

A setup similar to what is described in [Regulated Backward Compatibility and Deprecated Code](https://tribits.org/doc/TribitsUsersGuide.html#regulated-backward-compatibility-and-deprecated-code) would be good but apply it to TriBITS CMake code itself with respect to projects that use TriBITS.

What would likely be good would be to define and use a function like:

```
  tribits_deprecated(<msg>)
```

which would be called in any deprecated TriBITS function or macro.  By default, this would call:

```
  message(DEPRECATION <msg>)
```

but when `TRIBITS_HIDE_DEPRECATED_TRIBITS_WARNING_MESSAGES=TRUE`, then no deprecated warnings would be printed at all.

e10harvey:
I think having a way to enable/disable deprecation warning or error messages is sufficient. I would suggest adding one function for this purpose:
```python
tribits_deprecated(msg):
  if CMAKE_ERROR_DEPRECATED
    message(FATAL_ERROR, msg)
  else if CMAKE_WARN_DEPRECATED
    message(WARN, msg)
  else
    return
```

bartlettroscoe:
@e10harvey, here is the idea demonstrated with the CMake script `deprecated_msg.cmake`:

```cmake
function(tribits_deprecated msg)
  if (NOT TRIBITS_HIDE_DEPRECATED_TRIBITS_WARNING_MESSAGES)
    message(DEPRECATION "${msg}")
  endif()
endfunction()

# Example deprecated TriBITS function

if (NOT TRIBITS_HIDE_TRIBITS_DEPRECATED_CODE)

function(tribits_deprecated_function)
  tribits_deprecated("TriBITS function '${CMAKE_CURRENT_FUNCTION}()' is deprecated!")
endfunction()

[...]

PR #552, opened 2022-12-15
Complete deprecation work (#429)

bartlettroscoe:
### Description

There was a good bit of work left over from #429 that I noticed after I updated the snapshot of TriBITS in Trilinos and saw a bunch of deprecation warnings.  I also saw that TriBITS examples were using deprecated code as well.

The main commit in this PR is https://github.com/TriBITSPub/TriBITS/commit/69fb01c520c454fb2fe049aadf21263a75b2358c which updates the documentation and fixes some bugs.

This PR also does some other things (see the individual commits for details).

### Instructions for reviewers

Please review the commits one-by-one.  That will make the review the easiest to do and provide the best feedback.


bartlettroscoe:
Hello @KyleFromKitware, can you please give this PR a quick review?  I would like to get this PR merged as part of the TriBITS update of Trilinos in https://github.com/trilinos/Trilinos/pull/11380.

refactoring/token-replace.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def getCmndLineOptions():
+  from argparse import ArgumentParser, RawDescriptionHelpFormatter
+
+  clp = ArgumentParser(description=usageHelp,
+    formatter_class=RawDescriptionHelpFormatter)
+
+  clp.add_argument(
+    "-t", dest="tokenToReplace", required=True,
```
KyleFromKitware, line 21: May want to validate that this argument is a valid token (`^[a-zA-Z0-9_]+$`).

refactoring/token-replace.py
```diff
@@ -0,0 +1,77 @@
...
+#
+
+if __name__ == '__main__':
+
+  import sys, re
+
+  inOptions = getCmndLineOptions()
+
+  beginLineTokenPattern = re.compile(
+    r'^' + inOptions.tokenToReplace + r'([^A-Za-z0-9_])' )
```
KyleFromKitware, line 55: Using `f''` syntax may make this more readable.

refactoring/token-replace.py
bartlettroscoe: If I add unit tests for these simple tools, I will keep that in mind.  One would always carefully examine the impact of running this tool to replace tokens by running `git diff` to make sure it did what you expect.

refactoring/token-replace.py
bartlettroscoe: Is it possible to mix `r'...'` and `f'...'` in the same sting?

bartlettroscoe: FYI: I did a direct push to 'master' for this fix in commit https://github.com/TriBITSPub/TriBITS/commit/28a5c830dd51c488ca7ee299ff5fa6239f814c39.

There is no automated testing for that code so there is no point in creating a PR and running all of those GHA builds.

Issue #554, opened 2022-12-20
RUN_SERIAL missing from tribits_add_executable_and_test()


bartlettroscoe:
As described in trilinos/Trilinos#11411, people assumed that `tribits_add_executable_and_test()` supports the `RUN_SERIAL`

PR #555, opened 2022-12-20
Add RUN_SERIAL to tribits_add_executable_and_test() (#554)

bartlettroscoe:
Addresses #554.

It was an oversight that this was not added to tribits_add_executable_and_test() when it was added to tribits_add_test().

I tested this against Trilinos and the warning reported in https://github.com/trilinos/Trilinos/issues/11411 went away.

### DIFF
--- a/test/core/TestingFunctionMacro_UnitTests.cmake
+++ b/test/core/TestingFunctionMacro_UnitTests.cmake
@@ -4189,6 +4189,7 @@ function(unittest_tribits_add_executable_and_test)
     KEYWORDS keyword1 keyword2
     DEFINES -DSOMEDEFINE2
     TARGET_DEFINES -DSOMEDEFINE1
+    RUN_SERIAL
     ADD_DIR_TO_NAME
     LINKER_LANGUAGE C
     NUM_MPI_PROCS numProcs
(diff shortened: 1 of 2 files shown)


rabartlett1972 (approved):
I will allow to merge.  This does not need a post-merge review because the changes are trivial.

bartlettroscoe:
Also, I inspected the generated Trilinos file `packages/stokhos/test/UnitTest/CTestTestfile.cmake` and saw:

```
set_tests_properties(Stokhos_TpetraMatVec_MPI_4 PROPERTIES  ... RUN_SERIAL "ON" ...)
```

which is created from:

```
IF (Stokhos_ENABLE_Tpetra)
  TRIBITS_ADD_EXECUTABLE_AND_TEST(
    TpetraMatVec
    SOURCES tpetra_mat_vec.cpp
    COMM serial mpi
    STANDARD_PASS_OUTPUT
    #NUM_MPI_PROCS 4
    RUN_SERIAL
    )
ENDIF()
```

This validates that the `RUN_SERIAL` option was passed to the `tribits_add_test()` function correctly and that it was interpreted correctly by that function to set the `RUN_SERIAL` ctest property.

Issue #557, opened 2022-12-24
Updated TriBITS is disabling downstream TPLs


bartlettroscoe:
### Description

The updated TriBITS snapshotted to Trilinos in PR https://github.com/trilinos/Trilinos/pull/11380 has broken backward compatibility in that disabling an upstream TPL like HDF5 will disable a downstream TPL like NetCDF.  Well, that breaks some configure scripts for Trilinos like the one for Albany as reported in https://github.com/trilinos/Trilinos/issues/11426.

We, it turns out that I knew that was going to happen as I documented it in the [tribits/CHANGLOG.md 2022-10-20 entry](https://github.com/bartlettroscoe/Trilinos/blob/bfb7dc8f7ae500054be09074047a0a7486a31655/cmake/tribits/CHANGELOG.md#2022-10-20):

> **Changed:** Disabling an external package/TPL will now disable any downstream external packages/TPLs that list a dependency on that external package/TPL through its `FindTPL<tplName>Dependencies.cmake` file. Prior to this, disabling an external package/TPL would not disable dependent downstream external packages/TPLs (it would only disable downstream dependent required internal packages). To avoid this, simply leave the enable status of the upstream external package/TPL empty "" and no downstream propagation of disables will take place.

which you can see got posted in the PR in the file [tribits/CHANGLOG.md](https://github.com/trilinos/Trilinos/pull/11380/files#diff-344da4816d49ad8e76548f2548091cfe8be8af04a6f300463273473b12c12b20).

### Proposed solution

I think the proposed solution, for now, is to make TPL dependencies optional instead of required.  Such dependencies already behave like they are optional in that enabling a TPL like NetCDF will not automatically enable a dependent upstream TPL like HDF5.  To provide maximum flexibility in how people define TPLs, we should likely make all TPL dependencies optional.

NOTE: The reason that TPL dependencies had to be added was to correctly propagate and order TPL libraries to downstream targets.  TPL dependencies were not added to change enable/disable behavior.  (Not sure why I decided to make TPL dependencies behave this way.)

bartlettroscoe:
This was resolved with the merge of the PR:

* #558

PR #558, opened 2023-01-06
Make TPL-to-TPL dependencies optional (#557)

bartlettroscoe:
Disabling the upstream TPL HDF5 was disabling the TPL Netcdf and that broke several customer's Trilinos configure scripts.  The issue is, that we can't treat TPL-to-TPL dependencies as required for cases like this (see https://github.com/trilinos/Trilinos/issues/11426).

In the future, we may need to expand TPL intra-dependencies to support optional and required dependencies because there are cases where you would like the behavior of required dependencies for some TPL relationships.


rabartlett1972 (approved):
Letting this merge so I can snapshot to Trilinos.  Then I will set up a post-merge review.

bartlettroscoe:
Hello @KyleFromKitware, can you please do a post-merge review of this PR when you get a chance?

PR #564, opened 2023-02-23
tribits_add_option_and_define(): Error out if non-cache vars already defined (trilinos/Trilinos#10355)

bartlettroscoe:
See the commit message for details and:

* trilinos/Trilinos#10355
* trilinos/Trilinos#11583


bartlettroscoe:
@KyleFromKitware, here is a small one to review.  Should not take longer than 5 mins

KyleFromKitware (approved):

Issue #411, opened 2021-08-30
EPIC: Refactoring TriBITS and TriBITS-based CMake projects to confirm to modern CMake standards


bartlettroscoe:
While reading the book "Professional CMake

bartlettroscoe:
FYI, the non-usage of the standard CMake `FindMPI.cmake` is an issue that should be addressed as well.  See:

* https://github.com/sandialabs/Albany/issues/985#issuecomment-1738020649

bartlettroscoe:
FYI: It is amazing the shockwaves that have been produced in switching Trilinos to just use `GNUInstallDirs.cmake` by default  on even just the local the Trilinos user community.  See:

* Albany: https://github.com/sandialabs/Albany/issues/985
* Nalu: https://github.com/trilinos/Trilinos/issues/1232
* Sierra: https://github.com/trilinos/Trilinos/issues/12306#issuecomment-1737844074
* EMPIRE: Private email with Roger P.

SIDENOTE: It is amazing how much of a shockwave this one seemingly small non-backward compatible change to adopt one modern CMake practice has made on even just the local the Trilinos user community.  This does not bode well for all of the changes to switch to modern CMake listed above ☹️.

bartlettroscoe:
CC: @sebrowne, @ccober6 

NOTE: The TriBITS module `TribitsFindPythonInterp.cmake` (created by Bill Spotz back in the day?) currently calls `find_package(PythonInterp)` which uses the module `FindPythonInterp.cmake`.  According to:

* https://cmake.org/cmake/help/v3.23/module/FindPythonInterp.html

the `FindPythonInterp.cmake` module has been deprecated since CMake 3.12!  This should be updated to use `FindPython.cmake` or `FindPython3.cmake`.

PR #565, opened 2023-02-25
Update CMake min version from 3.17 to 3.23 (#411)

bartlettroscoe:
**Parent Issue:**

* #411

## Description

Addresses part of #411 and is part of:

* trilinos/Trilinos#10355

See last commit for more details.


bartlettroscoe:
FYI: The GHA builds shown [here](https://github.com/TriBITSPub/TriBITS/actions/runs/4267607142) for this PR branch shows that this works with the code as of the large PR #560.  Therefore, I will not rebase this one commit to be on top of the 'master' branch and deploy this.

NOTE: Any time you change the GHA builds, you have to manually change the set of required builds to pass a PR on the GitHub settings (which is a manual process).

bartlettroscoe:
Hello @KyleFromKitware, here is another smaller PR to review.  This needs to be reviewed and merged before the large PR #560 because I have already changed the setting of the branch protection for 'master' to require these updated GHA jobs for updated versions of CMake.

rabartlett1972 (approved):
I am going to allow the merge and then set up a post-merge review.

bartlettroscoe:
@KyleFromKitware, can you please do a post-merge reivew?

test/core/ExamplesUnitTests/TribitsExampleProject2_Tests.cmake
```diff
@@ -197,24 +197,24 @@ function(TribitsExampleProject2_find_tpl_parts  sharedOrStatic  findingTplsMetho
         "Tpl1_LIBRARY_NAMES='tpl1'"
         "Found lib '${tplInstallBaseDir}/install_tpl1/lib/libtpl1${libextregex}'"
         "TPL_Tpl1_LIBRARIES='${tplInstallBaseDir}/install_tpl1/lib/libtpl1${libextregex}'"
-        "Found header '${tplInstallBaseDir}/install_tpl1/include/Tpl1.hpp'"
+        "Found header '${tplInstallBaseDir}/install_tpl1/include[/]*/Tpl1.hpp'"
```
KyleFromKitware, line 200: Could you please give me a minimum `CMakeLists.txt` that reproduces the duplicate slash issue? This looks like a CMake bug to me, and I'd like to investigate. I tried a minimal `find_file(STDIO_H stdio.h)` and it didn't insert a duplicate slash.

Ignoring the broader issue, at the basic regex level, you could use `/+` instead of `[/]*/`.

bartlettroscoe: > Could you please give me a minimum CMakeLists.txt that reproduces the duplicate slash issue? This looks like a CMake bug to me, and I'd like to investigate. I tried a minimal find_file(STDIO_H stdio.h) and it didn't insert a duplicate slash.

Not sure how long that would take to create a minimal reproducer.  Is it not possible to just build a debug version of CMake and then directly debug the generation of this output?  You could even put in some debug print statements in the CMake C++ code to see what is happening.  That is likely the fastest way to track this down.

> Ignoring the broader issue, at the basic regex level, you could use `/+` instead of `[/]*/`.

Well, I would like to specifically like to match one or two `/` but not more than two.  But I guess you could use `/*/` and and that should work.

KyleFromKitware: `*` means "zero or more", not "zero or one". `/+` is the same thing as `[/]*/`. If you want "one or two", use `/?/` or `//?`.

bartlettroscoe: Addressed in PR #567.

Adrian-Diaz:
Is 3.23 a strict requirement for the Trilinos CMake config to work now? some of the systems I work with still only go up to 3.22 unfortunately.

bartlettroscoe:
> Is 3.23 a strict requirement for the Trilinos CMake config to work now?

Yes

> some of the systems I work with still only go up to 3.22 unfortunately.

Which systems, specifically?

PR #567, opened 2023-02-28
Replace regex used for handling multiple '/' from find_find() (#411)

bartlettroscoe:
This replaces '[/]*/' which will match one or more '/' chars in a row with '/?/' which will match one or two '/' in a row (which is what I wanted).

This addresses review feedback from @KyleFromKitware in PR #565.


rabartlett1972 (approved):
No need for a review. This implements suggestion in https://github.com/TriBITSPub/TriBITS/pull/565#discussion_r1119318668

PR #575, opened 2023-04-03
remove modern HDF5 version check

ibaned:
HDF5 actually uses an equality
function where major and minor
numbers must match.
This means the code was requiring
1.10 and 1.11 or higher would not
work.
It seems the best solution to just
remove this check altogether
since even the 1.10 version was
a guess

### DIFF
--- a/tribits/common_tpls/FindTPLHDF5.cmake
+++ b/tribits/common_tpls/FindTPLHDF5.cmake
@@ -6,10 +6,9 @@ set(HDF5_INTERNAL_IS_MODERN FALSE)
 
 if (Netcdf_ALLOW_MODERN)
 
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
> My feeling is that since this is all currently opt-in, it can be the user's responsibility to provide compatible versions of NetCDF and HDF5 and not TriBITS' responsibility to do any additional checking, but I support anything that doesn't prevent me from using newer HDF5 versions

Fair enough.  In this transition stage of the CMake package ecosystem to modern CMake, things are going to be messy along the way.  (Just hope we can get the part of the CMake ecosystem we care about updated to modern CMake before CMake gets replaced by Meson 😀)

Issue #577, opened 2023-05-08
cdash_analyze_and_report.py: HTML special characters in (e.g.,) build names are not escaped when generating the report


arghdos:
I have been experimenting with `cdash_analyze_and_report.py` to provide better reporting of my `CDash` project, and have got things mostly working.  One thing I have noticed however, is that certain links in the generated HTML file are not escaped properly, for example:

![image](https://user-images.githubusercontent.com/6463881/236906170-c65bee58-c0ca-4db7-9194-e5722c7f88b3.png)

Because I named my build "Linux-<sometestsuite>-with-<somecompiler>", I often have a "++" as a part of the build name.  When CDAR generates the links for such a build, you land on a filter page like e.g.:

![image](https://user-images.githubusercontent.com/6463881/236906518-0d7ca789-5377-4bbe-a46d-61ce3b77bc39.png)

I.e., the offending special characters ('++' in this case, but could be any HTML protected characters generally) are not displayed, and you get a filter that displays no builds.  

I think solving this would be as simple as using something like [urllib.parse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote) to escape the user provided strings (i.e., filterFields, and friends) in the various [QueryUrl constructors](https://github.com/TriBITSPub/TriBITS/blob/46dff4f893f3a4cac899fa2065c52c1349e3ad37/tribits/ci_support/CDashQueryAnalyzeReport.py#L725).

This is my first contribution to TriBITS, so please let me know if I've missed any of the guidelines.  If I'm on the right track here, I'm happy to work on a fix / tests, and submit a PR.

bartlettroscoe:
Hello @arghdos, yes, that is a problem in general.  If you can post a PR that fixes this using standard Python 3.x, then that would be great!

And if you have other suggestions for this tool, let me know.  (I have a long list of features that would make this tool more useful for various use cases that I should post to a GitHub issue.)

bartlettroscoe:
> Will do! If you write up your list of features, send me a link.

@arghdos, I did a dump into the new epic:

* #578

Also note there are a number of existing Issues shown in [this query](https://github.com/TriBITSPub/TriBITS/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+cdash_analyze_and_report.py+).

[...]

PR #579, opened 2023-05-09
Implement URL normalization in cdash_analyze_and_report.py (#577)

arghdos:
This commit:

1. Moves all URL normalization to use `urllib.parse.quote` from the [standard library](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote).  Note: I did add a python2 fallback, which I believe _should_ work, but I have not tested, nor do I know if that's a target at this time.
2. Adds project name URL normalization to all URL Query creator methods (and adds appropriate tests, see: `*_project_name_ampersand`).  Previously this was only applied in `getCDashQueryTestsQueryUrl`, but I felt it was best to be consistent.
3. Selectively applies URL normalization to the project, site, build and test ids (not 100% sure the last is necessary, but doesn't hurt) to the values used in `getCDashQueryTestsBrowserUrl` to construct the query URLs.  The quote function cannot be applied to the entire URL (which would be vastly more convenient) because, from the docs:

>By default, this function is intended for quoting the path section of a URL. The optional safe parameter specifies additional ASCII characters that should not be quoted — its default value is '/'.

That is, if you apply it to the full URL it will escape things you don't want it to (e.g., the '&' in the php queries).  Since we are manually constructing these, it is simple enough to apply.

4. Adds a note about why we _don't_ have to escape the queries coming from the input options (they are assumed to be normalized, as they're generated by CDash), and
5. Adds a corresponding test  (e.g., build name 'build++', site name with a space)

### DIFF
--- a/test/ci_support/CDashQueryAnalyzeReport_UnitTests.py
+++ b/test/ci_support/CDashQueryAnalyzeReport_UnitTests.py
@@ -982,6 +982,14 @@ class test_CDashQueryAnalyzeReport_UrlFuncs(unittest.TestCase):
       "&filtercount=1&morestuff"
     self.assertEqual(cdashIndexQueryUrl, cdashIndexQueryUrl_expected)
 
+  def test_getCDashIndexQueryUrl_project_name_ampersand(self):
+    cdashIndexQueryUrl = getCDashIndexQueryUrl(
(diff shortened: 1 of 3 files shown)


bartlettroscoe (approved):
The changes and the tests look great!  I should have now there was a standard Python library function like `urlquote()`.  (I think I did some looking at the time but I must not have found this.)

PR #588, opened 2023-06-30
gitdist: Pass in '-c color.status=always' when --dist-no-color is not added

bartlettroscoe:
Have `--dist-no-color` add `-c color.status=never` before very git repo command and set `-c color.status=always` otherwise.

### DIFF
--- a/tribits/python_utils/gitdist.py
+++ b/tribits/python_utils/gitdist.py
@@ -403,13 +403,13 @@ lines per repo in the file).  A compatible repo version file can be generated
 with this script listing three lines per repo (e.g. as shown above) using (for
 example):
 
-  $ gitdist --dist-no-color log -1 --pretty=format:"%h [%ad] <%ae>%n%s" \
+  $ gitdist --dist-no-color log --color=never -1 --pretty=format:"%h [%ad] <%ae>%n%s" \
     | grep -v "^$" &> RepoVersion.txt
 
 (which is defined as the alias 'gitdist-repo-versions' in the file
 'gitdist-setup.sh') or two lines per repo using (for example):
 
-  $ gitdist --dist-no-color log -1 --pretty=format:"%h [%ad] <%ae>" \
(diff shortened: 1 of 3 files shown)


bartlettroscoe:
@jmgate, not sure if you are still using gitdist or not but I don't know who else to ask to review a PR like this.

rabartlett1972 (approved):
I will go ahead and let this merge without outside review.

jmgate:
> @jmgate, not sure if you are still using gitdist or not&hellip;

I haven't used it in years, but I was about to pick it up again because I was just put on a project with dozens upon dozens of repositories.  I'll take a quick look.

tribits/python_utils/gitdist.py
```diff
@@ -1312,7 +1312,10 @@ def getCommandlineOps():
 
   clp.add_option(
     noColorArgName, dest="useColor", action="store_false",
-    help="If set, don't use color in the output for gitdist (better for output to a file).",
+    help="If set, don't use color in the output for gitdist and set"
+    +" '-c color.status=never' before the git command (like 'status')."
+    +"  NOTE: user should also pass in --color=never for git commands "
+    +" accept that argument.  (Better for output to a file).",
```
jmgate, line 1318:
```
for git commands that accept that argument.
                 ^^^^ missing?
```

jmgate (commented):
Might be worthwhile to add some smarts at some point to automatically add `--color=always/never` for the commands that accept it, so the user doesn't need to double-specify whether they want colors.

Also, do the `-c foo=bar` and `--color=foo` options require a certain version of git?  Probably doesn't matter, but thought it might be worth checking.

bartlettroscoe:
@jmgate, thanks for taking a look!

tribits/python_utils/gitdist.py
bartlettroscoe: I wanted to make the script as future proof as possible so not make it know too much about Git.  Currently, gitdist only knows a few Git commands in order to provide repo states and do a few other things.

PR #589, opened 2023-06-30
Fix: gitdist: dist-repo-status: Display tag or SHA1 instead of 'HEAD'

bartlettroscoe:
There was a bug is the previous PR #587.  See commit messages for details.

PR #587, opened 2023-06-29
gitdist: dist-repo-status: Display tag or SHA1 instead of 'HEAD'

bartlettroscoe:
The output is much more useful when in a detached head state.

Produces something like:

```
$ gitdist-status 
----------------------------------------------------------------------------------------------------
| ID | Repo Dir                         | Branch                    | Tracking Branch  | C | M | ? |
|----|----------------------------------|---------------------------|------------------|---|---|---|
|  0 | Trilinos (Base)                  | 11976-package_source-fix2 |                  |   |   |   |
|  1 | packages/framework/GenConfig     | 924a08a                   |                  |   |   |   |
|  2 | packages/framework/son-ini-files | master                    | glex-tril/master |   |   |   |
|  3 | packages/framework/srn-ini-files | master                    | origin/master    |   |   |   |
|  4 | TriBITS                          | gitdist-head-tag-and-sha1 |                  |   | 1 |   |
|  5 | kokkos                           | 4.1.00                    |                  |   |   |   |
|  6 | kokkos-kernels                   | 4.1.00                    |                  |   |   |   |
----------------------------------------------------------------------------------------------------
```

instead of:

```
$ gitdist-status 
--------------------------------------------------------------------------------------------------------------------
| ID | Repo Dir                         | Branch                    | Tracking Branch                  | C | M | ? |
|----|----------------------------------|---------------------------|----------------------------------|---|---|---|
|  0 | Trilinos (Base)                  | 11976-package_source-fix2 |                                  |   |   |   |
|  1 | packages/framework/GenConfig     | HEAD                      |                                  |   |   |   |
|  2 | packages/framework/son-ini-files | master                    | glex-tril/master                 |   |   |   |
|  3 | packages/framework/srn-ini-files | master                    | origin/master                    |   |   |   |
|  4 | TriBITS                          | gitdist-head-tag-and-sha1 | github/gitdist-head-tag-and-sha1 |   |   |   |
|  5 | kokkos                           | HEAD                      |                                  |   |   |   |
|  6 | kokkos-kernels                   | HEAD                      |                                  |   |   |   |
--------------------------------------------------------------------------------------------------------------------

```

PR #590, opened 2023-06-30
Fix: gitdist: dist-repo-status: Display tag or SHA1 instead of 'HEAD'

bartlettroscoe:
There was a bug is the previous PR #587.  See commit messages for details.

This replaces the following PR that seems to show a GitHub defect described in https://github.com/TriBITSPub/TriBITS/pull/589#issuecomment-1615052483:

* #589 

I was afraid that the GitHub defect would cause the merge of the wrong commits.


rabartlett1972 (approved):
I will let this PR merge without an external review.

PR #605, opened 2024-02-15
CDash Random failure tool patch 2024-02-15 (#600)

achauphan:
## Related issues
- #600 

## Description
These are requested changes to the [Trilinos production version](https://github.com/achauphan/TriBITS/pull/2) that are being implemented. I've packaged them as a patch here for the TriBITS version. 

Let me know if there is a better way to going about these patches going forward. I figured it would be a good idea to include the changes in the production Trilinos version of the script here as well.

## Changes

- Added optional `--email-subject-prefix` argument to add a prefix string to html page title and email subject line
- Change test history URL output in `singleSummaryReporter` to use a hyperlink to clean up output


bartlettroscoe (approved):
Changes look great and make perfect sense.

PR #608, opened 2024-06-06
Use 'python3' to access Python

### LINKED ISSUE #607: TriBITS Python scripts do not work on some modern Linux distributions
As per https://peps.python.org/pep-0394, `python` may not exist on some modern Linux operating systems.  The particular one in question here is RHEL8.  If you install one of the python3 packages, you do get `python3`, but not `python`.  As per the PEP, I think it wise to discontinue use of `python`, given the end-of-life of Python 2.

This specifically applies to the use of the `#!/usr/bin/env python` idiom present in many of the TriBITS Python scripts.  This is causing issues for the Trilinos AutoTester on RHEL8.

sebrowne:
As per https://peps.python.org/pep-0394, it is best practice to not use `python` in shebangs. Use the explicit `python3` entry point instead.

Closes #607 

I did not test this manually, I was planning on seeing how it works within the automated tests.

This is technically a break in backwards compatibility, but may not be experienced given how long Python 2 has been out of support.


refactoring/string-replace.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python3
+#!/bin/env python33
```
bartlettroscoe, line 1: This should be 'python3', no?

refactoring/token-replace.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python3
+#!/bin/env python33
```
bartlettroscoe, line 1: Again, this should be `python3`.

bartlettroscoe (approved):
I rebased the commits and remove the changes to the refactoring/ scripts that changed `python3` to `python33`. 

This goes well with the recent PRs:

* #616
* #617 

and think think this resolves #610.

Thanks @sebrowne!

Issue #612, opened 2024-07-30
TriBiTS must have version 3.23.0 or higher


zhengqigao:
When I do `./do-configure`, I get the following error:

```
Make Error at tribits/core/common/TribitsConstants.cmake:17 (message):
  Error, TriBiTS must have version 3.23.0 or higher!
Call Stack (most recent call first):
  tribits/core/package_arch/TribitsProjectImpl.cmake:33 (tribits_asesrt_minimum_cmake_version)
  tribits/core/package_arch/TribitsProject.cmake:40 (include)
  tribits/TriBITS.cmake:27 (include)
  CMakeLists.txt:14 (include)
```

What might be the problem?

zhengqigao:
3.18.0

zhengqigao:
Actually, I am installing Trilinos using Cmake. I manually change the CmakeList.txt in Trilinos to allow Cmake 3.18.0 and then compile Trilinos. But then it shows error:

```
cmake -DCMAKE_INSTALL_PREFIX=./build ./Trilinos
CMake Error at cmake/tribits/core/common/TribitsConstants.cmake:17 (message):
  Error, TriBiTS must have version 3.23.0 or higher!
Call Stack (most recent call first):
  cmake/tribits/core/package_arch/TribitsProjectImpl.cmake:33 (tribits_asesrt_minimum_cmake_version)
  cmake/tribits/core/package_arch/TribitsProject.cmake:40 (include)
  cmake/tribits/TriBITS.cmake:27 (include)
  CMakeLists.txt:30 (INCLUDE)
```

I guess I need to install TriBiTS first, but I failed with the error provided at the beginning of this post.

bartlettroscoe:
> 3.18.0

@zhengqigao, a CMake version 3.23.0 or higher is needed to build TriBITS or any project that uses TriBITS.  It appears we may need to improve that error message some to make this more clear.  I will put in a PR that does that.

bartlettroscoe:
@zhengqigao, here is the PR that should make this error message more clear:

* #613

Thanks for raising this issue!

bartlettroscoe:
> Actually, I am installing Trilinos using Cmake. I manually change the CmakeList.txt in Trilinos to allow Cmake 3.18.0 and then compile Trilinos. But then it shows error:

[...]

PR #613, opened 2024-07-30
Provide better error message for tribits_asesrt_minimum_cmake_version() (#612)

bartlettroscoe:
Addresses #612

The current error messages seemed to suggest that the TriBITS version was too low.  But this is an error message about the version of CMake being too long.


rabartlett1972 (approved):
If it passes the GHA builds and tests, this should be fine.  (Very hard to mess up this code.)

NOTE: Technically, this code is untested.  Testing it would require configuring a TriBITS project with a version of TriBITS that is too old (or hacking the constant `TRIBITS_CMAKE_MINIMUM_REQUIRED`) but that seems like too much effort for something a simple as this.

PR #614, opened 2024-08-01
Various updates to TriBITS while getting to run on CEE RHEL8 machine

bartlettroscoe:
These are misc commits that I added to get TriBITS tests to run on a CEE RHEL8 machine that only has python3.  I used the AUE modules to create a dev env (super easy except for GNU-12.3.0 openmpi not working).

It seems that the test suite is running very slow on my old RHEL7 machine 'crf450' and that machine will be taken offline soon inside of SNL (because RHEL7 is being retired).  This test suite runs in about 5.5 minutes with ctest -j16 on my hpws workstation (with 16 cores).  That is slow (because it is using a mounted filesystem for everything) but not the end of the world.


rabartlett1972 (approved):
The only changes under tribits/ are changing from 'python' to 'python3' for the *.py drivers.

Issue #610, opened 2024-07-26
Switch from find_package(PythonInterp) to find_package(Python3)


bartlettroscoe:
CC: @sebrowne, @ccober6, @rppawlo 

**Parent Epic:**

* #411 

## Description

The TriBITS module `TribitsFindPythonInterp.cmake` (created by Bill Spotz back in the day) currently calls `find_package(PythonInterp)` which uses the module `FindPythonInterp.cmake`.  According to:

* https://cmake.org/cmake/help/v3.23/module/FindPythonInterp.html

the `FindPythonInterp.cmake` module has been deprecated since CMake 3.12!  This should be updated to use `FindPython.cmake` or `FindPython3.cmake`.

In fact, it seems that the old `FindPythonInterp.cmake` module can't find versions of Python 3.0 and greater (i.e. with the name `python3`).  This causes problems (e.g. see #607).

The new `FindPython3.cmake` module returns a variable called `Python3_EXECUTABLE` while the old `FindPythonInterp.cmake` module returned a variable called `PYTHON_EXECUTABLE`.  Therefore, you can't just swap calling `find_package(PythonInterp ...)` with `find_package(Python3 ...)`.  Doing so would break all of the existing configure scripts for TriBITS projects that are currently setting `-DPYTHON_EXECUTABLE=$(which python3)` and it would break all of the internal TriBITS code that keys off of `PYTHON_EXECUTABLE`.  So this would be a major break in backward compatibility (but the changes would be easy to absorb on both ends). 

NOTE: We could make it easier for users of TriBITS projects by putting in a check for a cache var `PYTHON_EXECUTABLE` and error out telling users that they must instead set `Python3_EXECUTABLE`.  Then, we could make it easier for TriBITS projects using `PYTHON_EXECUTABLE` in internal CMake code by setting it to an error value like `PYTHON_EXECUTABLE is no longer supported! Please use Python3_EXECTUABLE instead!`.  And we would put in a `variable_watch(PYTHON_EXECUTABLE)` to and execute and error if the project tried to set `PYTHON_EXECUTABLE`.

## Workarounds

The current workaround is to just have users manually set:

```
-DPYTHON_EXECUTABLE=$(which python3)
```

But that is not great.

bartlettroscoe:
NOTE: It appears that all of the Trilinos GenConfig PR builds are explicitly setting `-DPYTHON_EXECUTABLE=$(which python3)` (which is computed in a more verbose way).  See:

[...]

PR #616, opened 2024-10-09
Transition to find_package(Python3) (#610)

bartlettroscoe:
## Description

Addresses #610 

## Notes to reviewers

* Prefer to review this PR commit-by-commit
* There are some commits at the beginning that are not directly related to this change but were bundled with this PR to avoid extra overhead
* Please make sure and review the updated documentation and `CHANGELOG.md` files


sebrowne:
RHEL8 has a default Python version of 3.6.8, so TriBITS projects + Python support will not work there.  RHEL9 has a default Python version of 3.9.18 so that's a-okay.  We have a large number of RHEL8 platforms, so that has the potential to cause issues.  How strong is the requirement for 3.8+?

bartlettroscoe:
> RHEL8 has a default Python version of 3.6.8, so TriBITS projects + Python support will not work there. RHEL9 has a default Python version of 3.9.18 so that's a-okay. We have a large number of RHEL8 platforms, so that has the potential to cause issues. How strong is the requirement for 3.8+?

It is not a strong requirement at all.  Any version of Python 3.x should be fine (and even Python 2.7 is technically is still supported by the code).  It is just that even Python 3.8 is at end-of-support this month according to https://devguide.python.org/versions/.

I went ahead and downgrade this to 3.6.8 (see commit https://github.com/TriBITSPub/TriBITS/pull/616/commits/6fb7d5e6e078bf6b9d25e80e8c386c2000eb5723).

But note the these SNL RHEL8 machines seem to also have Python 3.12.3 installed by default under `/usr/bin/python3.12` as well:

```
$ /usr/bin/python3.12 --version
Python 3.12.3
```

I am wondering if we should not be more pro-active and push the version of Python3 being used and supported to a more recent version?  According to https://devguide.python.org/versions/, Python 3.12 should be supported through 2028.  Just a thought.

bartlettroscoe:
FYI: Note that on SNL RHEL8 machines, `find_package(Python3)` finds Python 3.12, not 3.6 as shown by the TriBITS configure output:

```
-- Found Python3: /usr/bin/python3.12 (found suitable version "3.12.3", minimum required is "3.6") found components: Interpreter 
-- Python3_EXECUTABLE='/usr/bin/python3.12'
```

So unless you explicitly configure with `-D Python3_EXECUTABLE=$(which python3)`, people will be using and testing with Python 3.12 anyway.

sebrowne:
While that is valid, that package is an optional (additional) installation.  So the question is "Should this support RHEL8 _default_ Python?"  I'm fine with the proposed approach now that you've noticed that Python 3.12 is installed on the RHEL8 machines that we care about the most.  It will make us install that package in our UBI8 containers as well, but we can do that.

sebrowne (approved):

bartlettroscoe:
> While that is valid, that package is an optional (additional) installation. So the question is "Should this support RHEL8 _default_ Python?" I'm fine with the proposed approach now that you've noticed that Python 3.12 is installed on the RHEL8 machines that we care about the most. It will make us install that package in our UBI8 containers as well, but we can do that.

I think this is really a decision for the Trilinos Leadership team and close Trilinos stakeholders.  TriBITS does not have much of a dependence on Python.  (You can build test your software with just raw CMake/CTest 3.23+.)  Python is just gravy for TriBITS.

I will go ahead and merge this.

rabartlett1972 (approved):
@sebrowne approved so I will allow to merge.

PR #619, opened 2024-10-11
Port test suite to cee rhel8 and add pre-push git hookCee rhel8 port

bartlettroscoe:
Just a few tweaks to get test suite passing on cee-build030 rhel8 machine and adding a pre-push hook to check for signed off commits


rabartlett1972 (approved):
If it passes, it is good to push

PR #620, opened 2024-10-11
Fix up git pre-push hook that checks for signed off commits 

bartlettroscoe:
This commit checks for commits that are not signed off and correct them.  The usage is pretty self documenting so I don't think much documentation is needed beyond this.

This turned out to be much more complicated that I would have imaged.   (Much harder than a pre-commit hook, for example.)

This script really needs to have automated tests (like we have for gitdist).


rabartlett1972 (approved):
Let it merge

NOTE: This needs to have automated tests going forward if others are going to use this

PR #622, opened 2024-11-01
Document support for alternative libs with tribits_tpl_find_include_dirs_and_libraries() REQUIRED_LIBS_NAMES

bartlettroscoe:
This adds documentation for how to provide a list of alternative lib names with the `tribits_tpl_find_include_dirs_and_libraries()` argument `REQUIRED_LIBS_NAMES`.  This has been a supported and tested feature for a long time.  But amazingly, it was not documented!

This PR was motivated by a question from a Trilinos user about how to update some `FindTPL<tplName>.cmake` find modules in Trilinos to look for alternative names as well (like look for `openblas` if you can't find `blas`).


rabartlett1972 (approved):
Allow merge so we can see the rendered documentation and then do a post-merge review.

PR #627, opened 2025-02-19
Fix GHA check failures 2025-02-19

bartlettroscoe:
This PR should fix the lingering GHA check failures that popped up recently.   This should allow the nightly GHA builds to start working again and should allow PRs like #624 to pass tests and merge.

In particular, the GHA container upgraded the version of codespell it was  using (to codespell 2.2.6-1) which trigged a bunch of more misspelled words (which is good).  Also, the COVERAGE tests started timing out at 1 minute on testing day 2025-01-09.  Not clear why that happened but don't have time to see why that is the case.

See commit messages for more details.


rabartlett1972 (approved):
Allow to merge

PR #631, opened 2025-03-28
Use CMake 4.0.0 in GHA testing (trilinos/Trilinos#13731)

bartlettroscoe:
Now that CMake 4.0.0 is tagged and officially released, use it in GHA testing.

Related to trilinos/Trilinos#13731


rabartlett1972 (approved):
Okay if passes GHA build

PR #632, opened 2025-03-28
Add install-cmake-binary.sh

bartlettroscoe:
This only works for Linux for now, but it would not be hard to get this to also work for other systems if needed.


rabartlett1972 (approved):
Okay to merge if passes builds.  The only thing that will be tested here is the build of the RST file.

PR #633, opened 2025-04-16
Update reference docs to reflect code default

sebrowne:
When investigating some behavior in Trilinos, I used the documentation to ascertain the default value of `${PROJECT_NAME}_DISABLE_ENABLED_FORWARD_DEP_PACKAGES`, and determined it to be inconsistent with the "real" setting.  Fix that here.


bartlettroscoe:
@sebrowne, it appears you are correct:

https://github.com/TriBITSPub/TriBITS/blob/06c8e1b84e0a6e96b8389df1bc78abee0c3fc8e0/tribits/core/package_arch/TribitsGlobalMacros.cmake#L218

It seems the default got changed in the commit ec1d0a2998c7522de22f76a514da8dcc369c624e (allowing an override) but the documentation did not get updated correctly somehow and may have even been added incorrectly in commit d2214dfebab4a08c8247b76eb41ca1079c6e372.

bartlettroscoe:
@sebrowne, what was the scenario where you hit this situation?

IMHO, automated testing processes should have this set to `${PROJECT_NAME}_DISABLE_ENABLED_FORWARD_DEP_PACKAGES=ON` as it makes them much more robust. (That was our experience with CASL VERA.)

sebrowne:
Here's the linked PR where I repair what is wrong in Trilinos:

https://github.com/trilinos/Trilinos/pull/13991

It was exactly as you said, but we were sometimes setting it and sometimes not.  Whether it should be set or unset long-term is a matter for future discussion, but we intend to be setting it on all CI configurations at the moment.

bartlettroscoe:
@sebrowne, I don't see where `${PROJECT_NAME}_DISABLE_ENABLED_FORWARD_DEP_PACKAGES=ON` is being set in:

https://github.com/trilinos/Trilinos/pull/13991/files

?

PR #641, opened 2026-01-30
Correctly splitting the --extra-cmake-options input.

djabaay:
Description:
The use of commandLineOptionsToList wraps the default .split() command which splits on whitespace, not commas.
Fixes issue #640.


bartlettroscoe (approved):
If this passes the tests, it should be fine to merge.

bartlettroscoe:
Hum, for some reason, I can't get the GHA jobs to trigger.  It says "Expected: Waiting for status to be reported" but I don't see anything happening.  I rebased the branch off of 'master' and forced pushed and that did not trigger the GHA jobs.  This must be some type of defect with GHA to explain this.  The GHA was disabled for a while and that might be related here.

I ran this branch locally and it passed.  (I was running in a container which is why I created #642.)   I will just force merge this.

PR #643, opened 2026-02-02
Correctly splitting the --extra-cmake-options input

bartlettroscoe:
Copy of #641 to try to get GHA jobs to fire off


rabartlett1972 (approved):
We don't have automated tests for this argument, but for the simple use cases it supported before this change, it should be fine.  (If not, we will add a test fix it.)

PR #645, opened 2026-04-28
Add support for setting LLVM_PROFILE_FILE unique for each test

bartlettroscoe:
This will allow running the tests in parallel but be able to access per-test coverage info.  This sets the env var for each test:

```text
export LLVM_PROFILE_FILE=<test-binary-dir>/<fullTestName>_%p.profraw
```

This matches what is being done in:

* https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11891

but this works with any version of CMake/CTest.

See updated documentation.


rabartlett1972 (approved):
This maintains backward compatibility unless you select the new option.  So this is safe to merge to 'master'.

